PR #57, opened 2025-01-16
working on AnnData only Landscape

cornhundred:


cornhundred:
The original scope for this task was to “AnnData only Landscape visualization” but we will reduce the scope to what was already implemented in this branch, which was making the image layer optional for the Landscape visualization. 

I'll confirm that the visualization notebooks are still working and pre-processing still works before merging.

PR #58, opened 2025-01-24
DEGA-178-image-quantification

jaspreetishar:


src/celldega/pre/image_quantification.py
```diff
@@ -0,0 +1,165 @@
...
+    plt.show()
+
+    enhancer = ImageEnhance.Contrast(stain_image)
+    stain_image_with_contrast = enhancer.enhance(contrast_factor)
+
+    # Plot the stain image after contrast adjustment
+    plt.figure(figsize=(10, 10))
+    plt.imshow(stain_image_with_contrast, cmap='gray')
+    plt.title(f"Stain Image with Contrast Factor {contrast_factor}")
+    plt.show()
```
huanlity, line 38: I saw a few `plt.show()` in this file. If it is standard QC results plot, it is better to write to a file. If this function is called multiple time, you may want to consider skipping the `plt.show()` with `plt.close() or add a bool argument to switch on/off displaying plot.

src/celldega/pre/image_quantification.py
```diff
@@ -0,0 +1,165 @@
...
+    plt.figure(figsize=(10, 10))
+    plt.imshow(stain_bright_regions, cmap='hot')
+    plt.title(f"Filtered Bright Regions (Threshold: {intensity_threshold})")
+    plt.show()
+
+    tifffile.imwrite(f"{base_path}/contrast_adjusted_stain_image.tif", stain_bright_regions.astype(np.uint8))
+
+    return f"{base_path}/contrast_adjusted_stain_image.tif"
+
+def process_inputs_for_quantification(technology_name, cell_boundary_file, image_file, transform_file, base_path, image_contrast_required=False, contrast_factor=5, intensity_threshold=100):
```
huanlity, line 53: Some docstring below the function name can be very helpful here.

src/celldega/pre/image_quantification.py
```diff
@@ -0,0 +1,165 @@
...
+            except ValueError:
+                mean_intensity, sum_intensity = 0, 0
+
+            intensities.append(mean_intensity)
+            sum_intensities.append(sum_intensity)
+
+    cell_polygons["mean_intensity_EdU"] = intensities
+    cell_polygons["sum_intensity_EdU"] = sum_intensities
+
+    log_transformed_EdU_intensities = np.log1p(cell_polygons['sum_intensity_EdU'])
```
huanlity, line 104: Why `log1p` is used here and not other normalization methods?

src/celldega/pre/image_quantification.py
```diff
@@ -0,0 +1,165 @@
...
+    plt.title("Overlay of filtered stain-positive cells on stain image", fontsize=40)
+    plt.xticks(fontsize=40)
+    plt.yticks(fontsize=40)
+    plt.show()
+
+    filtered_cells.to_csv(f'{base_path}/filtered_stain_positive_cells.csv')
+
+    print("Filtered stain positive cells data (.csv) saved. Filtering done.")
+
+def image_quantification(base_path, technology_name, image_contrast_required=False):
```
huanlity, line 146: Do you have an example notebook or something that showcases this function? Would like to see an example. Thanks!

src/celldega/pre/image_quantification.py
jaspreetishar: I used log1p normalization because the image data I was working with had very low intensities for a specific stain, making the distribution highly skewed. I chose log1p to reduce the impact of extreme values and bring the data closer to a normal distribution. Do you think this is the right normalization method? If so, do you have any suggestions for a better normalization method?

src/celldega/pre/image_quantification.py
jaspreetishar: Just sent you an example notebook and an example dataset!

src/celldega/pre/image_quantification.py
```diff
@@ -0,0 +1,339 @@
...
+import glob
+import zarr
+import tifffile
+import pandas as pd
+import numpy as np
+import geopandas as gpd
+import matplotlib.pyplot as plt
+import rasterio
+from rasterio.mask import mask
+from rasterstats import zonal_stats
```
huanlity, line 11: I believe `zontal_stats` is never used, can you remove it along other functions that not being used in this file? Thx!

src/celldega/pre/image_quantification.py
```diff
@@ -0,0 +1,339 @@
...
+    transformed_cells = gpd.GeoDataFrame({'cell_index': merged_gdf['cell_index']}, geometry=merged_gdf['polygons'])
+    transformed_cells.to_parquet(f"{base_path}/transformed_cell_polygons.parquet")
+
+    if image_contrast_required:
+        contrast_adjusted_image_file = image_contrast_adjustment(base_path=base_path, technology_name=technology_name, contrast_factor=contrast_factor, intensity_threshold=intensity_threshold)
+        return (transformed_cells, contrast_adjusted_image_file)
+    else:
+        return transformed_cells
+    
+def calculate_stain_intensities(image_file, cell_polygons, technology_name):
```
huanlity, line 149: I imagine this would be an routinely used function based on my experience. Let's refactor it a a little bit and build a robuster version of it, meaning:

1. Break the function into a few parts, for reproducibility and easier maintenance. 
- `load_stain_img(fname, tech)` which returns `img`(array)
- `img_zonal_stats(input_img, seg_mask_or_seg_gdf)`
- `plot_zonal_stats()`

2. the `img_zonal_stats()` takes img as input layer instead of file names so it can be used in other places. Similarly, the mask can be an np array with the same size of `input_img`, but also can be a `gdf` (for flexibility), only in this case, you want to make sure the geometry of the `gdf` is in pixel space and match `input_img`.

3. The `log_transformed_stain_intensities` does not need to be returned if an one-liner can calculate it, especially if you take the suggestion and have a separate function for plotting.

@cornhundred please correct me if I am wrong here.

src/celldega/pre/image_quantification.py
```diff
@@ -0,0 +1,339 @@
...
+    counts, bin_edges, _ = plt.hist(log_normalized_stain_intensities, bins=50, edgecolor='black', alpha=0.7)
+    plt.title('Distribution of Log-Normalized Stain Intensities')
+    plt.xlabel('Log-Normalized Intensity Values')
+    plt.ylabel('Frequency of Occurrence')
+    plt.show()
+    plt.close('all')
+
+    return cell_polygons, log_transformed_stain_intensities, bin_edges
+
+def filtering_stain_positive_cells(highlighted_bin, bin_edges, log_transformed_stain_intensities, cell_polygons_with_metadata, base_path, technology_name, use_contrast_adjusted_image=False):
```
huanlity, line 254: It is indeed a `highlighted_bin` but what it actually is? The info behind matters so I propose something more direct like `positive_threshold` or something like this.

huanlity: If this is meant to be a manual and interactive tool for users to explore the right threshold for calling positive cells, we at lease need one more build-in argument to subset the image so a user can zoom in to the region of interest to investigate the effect of the threshold and fine-tune it.

I might be missing something but I am a little confused by the title of the two figures. Both of them have `stain-positive` in the title but the first one looks like all the cells to me and the second one looks like only the positive cells. Please clarify, thx!

Screenshot 2025-02-20 at 10 23 19 AM

src/celldega/pre/image_quantification.py
huanlity: You probably want to include `rasterio` in the toml file if we decide to use it.

huanlity:
@jaspreetishar Thanks for sending me the notebook and data for testing. They work fine on my end, except that I need to pip install `rasterio` on my end. I commented above to include it in the toml file.

After reviewing the script and the notebook, my biggest question for you is: what are you quantifying on the imagery data? The name of the ticket and and files suggested that but I am only seeing cell-level zonal stats calculation and filtering of cells based on manually inputted threshold. It is more like a interactive tool for positive calling? am I wrong about this? Thanks!

src/celldega/pre/image_quantification.py
jaspreetishar: I have added the functionality to allow testing on a subset of the image. Additionally, I have included print statements for both plots to clarify their contents. The first plot displays all cells with a stain intensity above zero, while the second plot shows only the cells with a stain intensity above a user-defined threshold.

jaspreetishar: Writing my reply here as I can't figure out how to reply directly to the latest comment:  

The imagery data is used to quantify stain-positive cells. The code indeed functions like an interactive tool designed to interactively identify and analyze stain-positive cells based on a user-defined threshold.

src/celldega/pre/image_quantification.py
```diff
@@ -0,0 +1,558 @@
...
+        zarr.Group: The root group of the opened Zarr dataset.
+    """
+
+    store = (zarr.ZipStore(path, mode="r")
+    if path.endswith(".zip")
+    else zarr.DirectoryStore(path)
+    )
+    return zarr.group(store=store)
+
+def load_stain_img(image_file, image_interval):
```
huanlity, line 36: `interval` is a little misleading, I suggest use `subset_bounds`.

src/celldega/pre/image_quantification.py
```diff
@@ -0,0 +1,558 @@
...
+
+    """
+
+    log_transformed_stain_intensities = np.log1p(cell_polygons['sum_stain_intensity'])
+
+    # Normalize the log-transformed intensities to the range [0, 1]
+    log_normalized_stain_intensities = (log_transformed_stain_intensities - log_transformed_stain_intensities.min()) / \
+                                     (log_transformed_stain_intensities.max() - log_transformed_stain_intensities.min())
+
+    print("Plotting cells with total stain intensity greater than zero:")
```
huanlity, line 365: If I understand correctly, `total `stain` here means the cumulative pixel number (or brightness values) within a cell, which seems depending on the size of a cell, while `mean` or `median` value of the intensities is more like a `average` representation of the expression at cell level. What is your thought?

Thanks!

huanlity:
@jaspreetishar thanks for the quick response and changes, please see my comments above. Everything else looks good to me.

src/celldega/pre/image_quantification.py
```diff
@@ -427,7 +399,7 @@ def calculate_stain_intensities(image_file, cell_polygons, technology_name):
 
     return cell_polygons, image_array, log_transformed_stain_intensities, bin_edges
 
-def filtering_stain_positive_cells(filtering_positive_cells_threshold, bin_edges, log_transformed_stain_intensities, cell_polygons_with_metadata, base_path, image_array, use_contrast_adjusted_image=False):
+def filtering_stain_positive_cells(filtering_positive_cells_threshold, bin_edges, log_transformed_stain_intensities, cell_polygons_with_metadata, base_path, image_array, subset_bounds, use_contrast_adjusted_image=False):
```
huanlity, line 402: This is optional but if you want to give the user a little bit more power to adjust the `vmin` and `vmax` to enhance the contrast to their liking, you can take advantage of the built-in arguments  `vmin` and `vmax` in `plt.imshow()` to do that. The `vmin` and `vmax` values can be automated using `np.quantile()` or something like that.
Screenshot 2025-02-24 at 2 16 50 PM

huanlity:
LGTM! ✅

cornhundred:
looks good @jaspreetishar, let me just have a look over and then we can merge if you are not making any more changes

jaspreetishar:
I have added the last of the requested changes. Please go ahead with the review/ merge

cornhundred (changes requested):
A couple of thoughts, can you add an example notebook? Then we can we work to visualizing the results in the Landscape view.

cornhundred (commented):
waiting to implement this

PR #59, opened 2025-01-25
DEGA-184-3D-pointcloud

cornhundred:


cornhundred:
There is a change to the `get_layers_list` arguments which required a lot of small changes to a set of files. The main large change is adding a point_cloud layer and a dependency on a new landscape_parameters.json key called landscape_type

PR #65, opened 2025-02-21
[DEGA-189] code for calculating spatial gradients and distributions with respect to regions

huanlity:
1. move functions from notebooks to a separate .py file, including calculating distance, calculate gradient (bands), and plotting respective results.
2. to make the code more modular, reusable, and efficient, I organized the functions in the new .py file as a class.

The code can be called like this:

```
# Specify the polygon
roi_name = 'region_1'

# Initialize the class
distance_calculator = DistanceToPolygon(gdf_micron, gdf_join_micron, roi_name)

# Calculate distances
distance_calculator.calc_distance_to_polygon()

# Plot distances
distance_calculator.plot_distance_to_polygon()

# Calculate gene gradients
band_width = 200
distance_calculator.calc_gene_gradient_to_polygon(band_width=band_width, num_bands=5, gene_list=gene_list)

# Plot gene gradients
distance_calculator.plot_gene_gradient_to_polygon(gene_list=gene_list, markersize=1)

# Save bands to Parquet
distance_calculator.save_bands_to_parquet(f"{path_landscape_files}/gdf_{band_width}_um_bands_from_{roi_name}.parquet")
```


cornhundred:
Hi @huanlity, I feel like I want to avoid making a class for this and we can instead use SpatialData (https://spatialdata.scverse.org/en/stable/tutorials/notebooks/notebooks/examples/tables.html) to store our state. Can you give that a try and let me know if you want to discuss?

Also, I want to use the Landscape visualization method to interactively visualize these regions and their associated gradients in addition to static plots. I think it is doable to add this to this pull request.

cornhundred (changes requested):
https://github.com/broadinstitute/celldega/pull/65#issuecomment-2683207382

huanlity:
Screenshot 2025-04-09 at 1 22 36 PM

cornhundred:
Hi @huanlity, I'm starting to review but the first thing that comes to mind is that this method should not live in the pre module since this is not a preprocess step, but we should make a spatial module for this method. 

The spatial module can also hold other methods that we will develop in the future like: spatial autocorrelation of genes within a cluster of cells, spatial autocorrelation of cell clusters within a tissue, calculation of genes that correlate with distance to a region, calculation of cell clusters populations that correlate with distance to a region.

huanlity:
I created a `sp` (spatial) module and moved functions there.

src/celldega/sp/__init__.py
```diff
@@ -0,0 +1,258 @@
...
+            - gdf_points: The input gdf_points with added distance column.
+            - gdf_polygon: The extracted polygon for the specified ROI.
+    """
+
+    print (f"Extract the polygon for the specified ROI: {roi_name}")
+    # Extract the polygon for the specified ROI
+    gdf_polygon = gdf_polygons.loc[gdf_polygons["roi"] == roi_name].reset_index()
+    
+    # Calculate distances from points to the polygon
+    gdf_points[f"distance_to_polygon_{roi_name}"] = gdf_points["geometry"].apply(
```
cornhundred, line 36: can you check if this can be vectorized to improve performance. Something like this `gdf_points.geometry.distance(target_geom)` maybe?

src/celldega/sp/__init__.py
```diff
@@ -0,0 +1,258 @@
...
+        lambda point: gdf_polygon.geometry.iloc[0].distance(point)
+    )
+    print (f"Distances from points to polygon {roi_name} were calculated.")
+    return gdf_points, gdf_polygon
```
cornhundred, line 40: I think we can no return `gdf_polygon` since this GeoDataFrame is not being updated.

src/celldega/pre/distance.py
cornhundred, line 1: Can we delete this file since it is the same as the __init__.py file under the spatial module?

src/celldega/sp/__init__.py
```diff
@@ -0,0 +1,258 @@
...
+    plt.title(
+        f"Region Selected Cells Colored Based on Distance to the Polygon {roi_name} (micron)"
+    )
+    plt.tight_layout()
+    plt.xlabel("x (micron)")
+    plt.ylabel("y (micron)")
+    plt.show()
+
+
+def create_concentric_rings(polygon, num_bands, band_width):
```
cornhundred, line 123: I think we should refactor create_concentric_rings out of this function and instead pass the output GeoDataFrame to this function.

src/celldega/sp/__init__.py
```diff
@@ -0,0 +1,258 @@
...
+        ring = outer.difference(inner)
+
+        band = polygon.copy()
+        band["geometry"] = ring
+        band["band"] = i
+        band_list.append(band)
+
+    return gpd.GeoDataFrame(pd.concat(band_list, ignore_index=True)).set_crs('EPSG:4326')
+
+def calc_gene_gradient_to_polygon(
```
cornhundred, line 154: Can we rename this to somethinig like calc_gene_expression_by_band? We're not identifying gradients with this function but rather calculating the average expression from cells that are within spatial bands.

But can we add a function called something like exp_corr_distance that calculates the correlation of genes with respect to their distance to an ROI? ChatGPT suggests something like this



```
def correlate_expression_with_distance_band(
    gdf: gpd.GeoDataFrame,
    gene_list: list,
    band_column: str = "band",
    method: str = "spearman"
) -> pd.DataFrame:
    """
    Calculate correlation between gene expression and distance bands.

    Args:
        gdf: GeoDataFrame with gene expression and distance band assignments.
        gene_list: List of genes to evaluate.
        band_column: Name of the column containing band identifiers.
        method: 'pearson' or 'spearman' correlation.

    Returns:
        DataFrame with columns: gene, correlation, p_value
    """
    from scipy.stats import spearmanr, pearsonr

    results = []
    for gene in gene_list:
        df = gdf[[band_column, gene]].dropna()
        if len(df) < 3:
            continue
        x = df[band_column]
        y = df[gene]
        corr_func = spearmanr if method == "spearman" else pearsonr
        corr, pval = corr_func(x, y)
        results.append({"gene": gene, "correlation": corr, "p_value": pval})

    return pd.DataFrame(results).sort_values(by="correlation", ascending=False)

```

src/celldega/sp/__init__.py
```diff
@@ -0,0 +1,258 @@
...
+"""
+
+
+import matplotlib.pyplot as plt
+import matplotlib.colors as mcolors
+import matplotlib.cm as cm
+import geopandas as gpd
+import pandas as pd
+
+def calc_distance_to_polygon(
```
cornhundred, line 12: Maybe we can start calling this ROI instead of polygon in our functions?

src/celldega/pre/__init__.py
```diff
@@ -952,4 +953,4 @@ def _check_required_files(technology, data_dir):
         )
 
 
-__all__ = ["landscape", "trx_tile", "boundary_tile"]
\ No newline at end of file
+__all__ = ["landscape", "trx_tile", "boundary_tile", "distance"]
```
cornhundred, line 956: we do not need to expose "distance" in the pre module as far as I know.

src/celldega/pre/__init__.py
```diff
@@ -34,6 +34,7 @@
 from ..clust import *
 from .image_info import *
 from .run_pre_processing import *
+from .distance import *
```
cornhundred, line 37: we can delete this I think

notebooks/BNB_Distance_to_A_Region_and_Gene_Expression_Gradient.ipynb
cornhundred, line 1: Could we modify this example to load data directly from a dataset instead of from the LandscapeFiles? We can first load raw Xenium data to SpatialData or AnnData and then we will automatically have cell positions in micron space. We probably do not want users to have to worry about pixel space since this will be handled by Celldega behind the scenes right before plotting.

notebooks/BNB_Metadata_and_Gene_Expression_of_Region_Selected_Cells.ipynb
cornhundred, line 1: Could you add more markdown explanation to the notebook? I'm not following completely.

notebooks/BNB_Neighborhood_with_Alphashape.ipynb
cornhundred, line 1: Similar to the comment about reading from the raw data rather than from the LandscapeFiles, can we set up something that will load the raw data (cell centers) in micron space, calculate alpha shapes in micron space, and then create a pixel space version of the alpha shapes for Celldega to plot? I'm thinking that we want to have the pixel space conversion take place under the hood instead of in a way where the user has to worry about having two versions of the alpha shapes. Maybe the conversion can take place in Python right before we pass the data as a traitlet to the front end? Let me know if you want to discuss.

notebooks/BNB_Region_Selection_and_Cells_Extraction.ipynb
cornhundred, line 1: Converting to micron space by default and pixel space for visualization only will simplify this notebook.

cornhundred (changes requested):
Hi @huanlity, thanks this is looking better. I made some suggestions. Please let me know if you want to go over anything.

src/celldega/sp/__init__.py
huanlity: fixed.

src/celldega/pre/distance.py
huanlity: My bad, thought I removed it already. fixed it now.

src/celldega/sp/__init__.py
huanlity: fixed.

src/celldega/pre/__init__.py
huanlity: fixed.

src/celldega/pre/__init__.py
huanlity: fixed.

notebooks/BNB_Distance_to_A_Region_and_Gene_Expression_Gradient.ipynb
huanlity: fixed.

huanlity:
Update the progress:

Screenshot 2025-04-17 at 8 45 00 PM

cornhundred:
@huanlity is that alpha shape plot showing all resolutions of alpha shape simultaneously?

src/celldega/sp/__init__.py
```diff
@@ -0,0 +1,359 @@
...
+    plt.title(
+        f"Region Selected Cells Colored Based on Distance to the ROI {roi_name} (micron)"
+    )
+    plt.tight_layout()
+    plt.xlabel("x (micron)")
+    plt.ylabel("y (micron)")
+    plt.show()
+
+
+def create_concentric_rings(polygon, num_bands, band_width):
```
cornhundred, line 128: It would be nice to not have to say how many bands we want but instead have this done automatically. One way to do this would be to first calculate the alpha shape of the entire tissue and then calculate a large number of rings (way more than we need, ~100), then only keep the rings that overlap with the alpha shape of the tissue (you could even take the intersection of the tissue-alpha-shape and ring if you want).

src/celldega/sp/__init__.py
```diff
@@ -0,0 +1,359 @@
...
+
+        ax.invert_yaxis()
+        plt.title(f"Gradient of {gene} Expression per Cell Along Distance to ROI")
+        plt.xlabel("x (micron)")
+        plt.ylabel("y (micron)")
+        plt.tight_layout()
+        plt.show()
+
+
+def create_band_pixel(gdf_micron, gdf_pixel, roi_name, n_rings, band_width, json_fname):
```
cornhundred, line 265: can we hide this from the user facing API (add underscore to name)?

cornhundred (changes requested):
I requested a few changes, but overall looking good.

src/celldega/sp/__init__.py
```diff
@@ -0,0 +1,359 @@
...
+        
+        # Apply to feature (scale 0-255)
+        feature['properties']['color'] = [
+            int(r * 255),
+            int(g * 255),
+            int(b * 255),
+            alpha
+        ]
+
+def landscape_ist_with_spatial_region(gdf_pixel, path_landscape_files, spatial_region_json_fname, color_metric, tech='Xenium', alpha=50):
```
cornhundred, line 325: Can we remove this function since we're going to come up with a cleaner method for plotting this in the Landscape view. We have this approach - https://broadinstitute.github.io/celldega/examples/brief_notebooks/Mouse-Brain_Alpha-Shape-Neighborhood/#calculate-alpha-shape-neighborhoods but we will improve this by hiding the GeoJSON from the user.

src/celldega/sp/__init__.py
```diff
@@ -0,0 +1,359 @@
...
+        color_metric: Metric for color assignment.
+        tech: Technology used (default is 'Xenium').
+        alpha: Alpha value for color transparency (default is 50).
+    Returns:
+        landscape_ist: A Landscape object with the spatial region and assigned colors.
+    """
+
+    # save alphashape json
+    gdf_pixel[color_metric] = gdf_pixel[color_metric].astype(int)
+    gdf_pixel.to_file(spatial_region_json_fname, driver='GeoJSON')
```
cornhundred, line 341: we do not need to save to a file and then load this file.

src/celldega/sp/__init__.py
cornhundred, line 128: One last thing - based on our new framework for how we're thinking about neighborhoods can we move the function that calculates the concentric bands and returns the GeoDataFrame into the `nbhd` module? We can also rename it from `create_concentric_rings` to `calc_grad_nbhd_from_roi` or something like this? Then the spatial module can contain

cornhundred (changes requested):
I added some additional comments.

src/celldega/nbhd/__init__.py
```diff
@@ -175,6 +175,10 @@ def alpha_shape_cell_clusters(meta_cell, cat='cluster', alphas=[100, 150, 200, 2
 
     gdf_alpha = gdf_alpha.loc[gdf_alpha.area.sort_values(ascending=False).index.tolist()]
 
+    # Explode MultiPolygon geometries into individual Polygons
+    gdf_alpha = gdf_alpha.explode(index_parts=True)
```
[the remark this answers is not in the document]
huanlity: fixed

src/celldega/sp/__init__.py
```diff
@@ -0,0 +1,359 @@
...
+    
+    Returns:
+        A GeoDataFrame containing the bands and mean gene expression.
+    """
+
+    print (f"Complete calculating gene exp. gradient from points to {roi_name} using {band_width} micron rings.")
+
+
+    # Merge all bands into a single GeoDataFrame
+    gdf_bands = create_concentric_rings(gdf_polygon, num_bands, band_width)
```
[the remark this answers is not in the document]
huanlity: fixed

src/celldega/sp/__init__.py
```diff
@@ -0,0 +1,359 @@
...
+    # 2. Create bands in pixel space (scaled by conversion factor)
+    gdf_pixel = gdf_pixel.loc[gdf_pixel["roi"] == roi_name].reset_index()
+    gdf_bands_pixel = create_concentric_rings(gdf_pixel, n_rings, band_width/conversion_factor)
+    
+    # 3. Save to GeoJSON
+    gdf_bands_pixel.to_file(json_fname, driver='GeoJSON')
+    return gdf_bands_pixel
+
+
+def assign_distinct_colors(spatial_region, color_metric='cat', alpha=100):
```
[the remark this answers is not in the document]
huanlity: fixed

src/celldega/sp/__init__.py
huanlity: fixed

src/celldega/sp/__init__.py
huanlity: this function is removed now

src/celldega/sp/__init__.py
huanlity: modified

huanlity: to reduce computing and the depandency on alphashapes, I will just compare with a gdf_reference which can be gdf of tissue cells that are readily available.

src/celldega/sp/__init__.py
```diff
@@ -0,0 +1,247 @@
...
+    
+    # Calculate distances from points to the polygon
+    polygon_geom = gdf_polygon.geometry.iloc[0]
+
+    gdf_points_with_dist = gdf_points.copy()
+    gdf_points_with_dist[f"distance_to_polygon_{roi_name}"] = gdf_points_with_dist.distance(polygon_geom)
+    print (f"Distances from points to polygon {roi_name} were calculated.")
+    return gdf_points_with_dist, gdf_polygon
+
+def plot_distance_to_roi(
```
cornhundred, line 47: I think going forward we can try to put all plotting methods under the viz module, but we can rearrange this later if needed.

src/celldega/sp/__init__.py
```diff
@@ -0,0 +1,247 @@
...
+    plt.tight_layout()
+    plt.xlabel("x (micron)")
+    plt.ylabel("y (micron)")
+    plt.show()
+
+
+def calc_gene_expression_by_band(
+    gdf_points: gpd.GeoDataFrame,
+    gdf_bands: gpd.GeoDataFrame,
+    roi_name: str,
```
cornhundred, line 131: can we remove the `roi_name` argument since it is not being used to perform any calculations.

src/celldega/sp/__init__.py
```diff
@@ -0,0 +1,247 @@
...
+    band_width: float,
+    gene_list: list
```
cornhundred, line 133: As far as I understand, `gdf_points` contains a subset of genes from the single cell data. Can we modify this so that we pass in as an argument the entire AnnData and/or a cbg matrix so that we calculate the average expression of all genes along these gradient bands?

notebooks/BNB_Distance_to_A_Region_and_Gene_Expression_Gradient.ipynb
cornhundred, line 1: Can we pass the entire AnnData instead of adding a subset of genes, `gene_list` to the GeoDataFrame for cell centroids like we do below? We want to be able to have the average expression along the gradient bands calculated for all genes.

```
# Load h5ad file
adata = sc.read_h5ad(f'{data_dir}.h5ad')
adata.obs.set_index('cell_id', inplace=True)

# Extract gene expressions foro a gene list (handles sparse/dense matrices)
gene_list = ['EPCAM', 'AMY2A', 'GATM']

gene_exp = pd.DataFrame(
    data=adata[:, gene_list].X.toarray() if hasattr(adata.X, 'toarray') else adata[:, gene_list].X,
    columns=gene_list,
    index=adata.obs_names
)

# Create combined DataFrame with geometry
gdf_cell = gpd.GeoDataFrame(
    data={
        'cluster': adata.obs['leiden'],
        **gene_exp  # Unpacks all gene columns
    },
    geometry=[Point(xy) for xy in adata.obsm['spatial'][:, :2]],
    crs="EPSG:4326"  # Set your coordinate system here
)
```

cornhundred: We can also drop the BNB from the names of these notebooks. We will later move a subset of them to the documentation.

cornhundred (changes requested):
Hi @huanlity, I didn't realize that the gene expression along the gradient was only being calculated for a subset of genes. Once we update it to take the AnnData as an argument and calculate the average expression along the gradient bands we should be good. I think the `BNB_Distance_to_A_Region_and_Gene_Expression_Gradient.ipynb` is the most relevant and we can rename it `Gene_Expression_Gradient_with_respect_to_ROI.ipynb`. We can also move it to the `docs/examples/brief_notebooks` directory so we can add it to the documentation.

src/celldega/sp/__init__.py
huanlity: fixed

src/celldega/sp/__init__.py
huanlity: fixed

src/celldega/sp/__init__.py
huanlity: fixed.

notebooks/BNB_Distance_to_A_Region_and_Gene_Expression_Gradient.ipynb
huanlity: fixed.

huanlity:
@cornhundred Thanks for the review, I addressed all the comments. Specifically, now we pass Andata to two functions: `calc_distance_to_roi()` and `calc_gene_expression_by_band()` and calculate the gradient for all available genes. I also moved the plotting functions to `viz` module.

cornhundred (commented):
waiting for update to NBHD methods

PR #66, opened 2025-02-26
DEGA-101-harmonize-segmentation-results

cornhundred:


src/celldega/nbhd/__init__.py
```diff
@@ -205,3 +213,190 @@ def alpha_shape_geojson(gdf_alpha, meta_cluster, inst_alpha):
...
+
+    gdf_trx = gpd.GeoDataFrame(trx, geometry=gpd.points_from_xy(trx[x], trx[y]))
+    gdf_trx['geometry_image_space'] = gdf_trx['geometry'].apply(lambda geom: affine_transform(geom, [transformation_matrix[0, 0],
+                                                                                               transformation_matrix[0, 1],
+                                                                                               transformation_matrix[1, 0],
+                                                                                               transformation_matrix[1, 1],
+                                                                                               transformation_matrix[0, 2],
+                                                                                               transformation_matrix[1, 2]]))
+
+    gdf_trx.set_geometry('geometry_image_space', inplace=True)
```
huanlity, line 333: The `unassigned_transcripts_tiled_view()` step is pretty low on my end. I guess the `apply` with `affine_transform` is the slow part as it is not a vectorized operation. Can you try something like below that I used in the codebase and that served me well? 

```
from ..pre.boundary_tile import batch_transform_geometries
from ..pre.__init__ import _to_geometry
gdf_trx["geometry"] = batch_transform_geometries(gdf_trx["geometry"], transformation_matrix, 1)
gdf_trx['geometry'] = gdf_trx['geometry'].apply(_to_geometry)
```

src/celldega/nbhd/__init__.py
```diff
@@ -205,3 +213,190 @@ def alpha_shape_geojson(gdf_alpha, meta_cluster, inst_alpha):
...
+
+    print("Calculating percentage of hexatile-specific unassigned transcripts...")
+
+    unassigned_counts = gdf_hexatile_assigned_trx[gdf_hexatile_assigned_trx['cell_id'] == "UNASSIGNED"].groupby('polygon_index').size()
+    total_counts = gdf_hexatile_assigned_trx.groupby('polygon_index').size()
+
+    percentage_unassigned = (unassigned_counts / total_counts) * 100
+    percentage_unassigned = percentage_unassigned.fillna(0)
+
+    percentage_unassigned = percentage_unassigned[percentage_unassigned < 75]
```
huanlity, line 357: Instead of hardwiring here, can you create an argument for value 75 so users can change if needed? Thx.

src/celldega/nbhd/__init__.py
```diff
@@ -205,3 +213,190 @@ def alpha_shape_geojson(gdf_alpha, meta_cluster, inst_alpha):
...
+    gdf_hexatile_assigned_trx.to_parquet(os.path.join(path_landscape_files, "hexatile_assigned_trx.parquet"))
+
+    print("Assignment of transcripts done and saved.")
+
+    print("Calculating percentage of hexatile-specific unassigned transcripts...")
+
+    unassigned_counts = gdf_hexatile_assigned_trx[gdf_hexatile_assigned_trx['cell_id'] == "UNASSIGNED"].groupby('polygon_index').size()
+    total_counts = gdf_hexatile_assigned_trx.groupby('polygon_index').size()
+
+    percentage_unassigned = (unassigned_counts / total_counts) * 100
```
huanlity, line 354: Consider doing the 2 groupby operations in one go to speed up:
```
counts = gdf_hexatile_assigned_trx.groupby('polygon_index')['cell_id'].agg(
    total='size',
    unassigned=lambda x: (x == "UNASSIGNED").sum()
)
percentage_unassigned = (counts['unassigned'] / counts['total']) * 100
```

src/celldega/nbhd/__init__.py
```diff
@@ -205,3 +213,190 @@ def alpha_shape_geojson(gdf_alpha, meta_cluster, inst_alpha):
...
+    return round(sum(gdf_hexatile.area.tolist())) == round(tissue_img_shape[0] * tissue_img_shape[1])
+
+def hexatile_trx_assignment(hexatile_assigned_trx):
+
+    more_than_one_matches = 'more_than_one_matches' in hexatile_assigned_trx['polygon_index'].unique()
+    unassigned = 'UNASSIGNED' in hexatile_assigned_trx['polygon_index'].unique()
+
+    return more_than_one_matches, unassigned
+
+def unassigned_transcripts_tiled_view(gdf_hexatile, path_default_data, path_landscape_files, path_segmentation_files=None):
```
huanlity, line 295: Some docstrings here can be very helpful.

src/celldega/pre/merge_segmentations.py
```diff
@@ -0,0 +1,514 @@
...
+        gdf_transcripts['polygon_index'] = gdf_transcripts.apply(find_containing_polygon,
+                                                                   axis=1,
+                                                                   polygons_sindex=polygons_sindex,
+                                                                   gdf_polygons=gdf_polygons)
+    else:
+        gdf_transcripts['polygon_index'] = 'UNASSIGNED'
+
+    return gdf_transcripts
+
+def assigning_transcripts(gdf_polygons, gdf_transcripts):
```
huanlity, line 64: After I looked deeper, I found that this function is the bottleneck, as it is the most time-consuming step. Current implementation processes transcripts one-by-one using `apply()`. I never finished processing it in the notebook. Please consider optimizing this.

An example can be (with the help of Deepseek):

```
from shapely.vectorized import contains
import numpy as np

def assign_transcripts(gdf_transcripts, gdf_polygons):
    """Optimized vectorized assignment"""
    polygons_sindex = gdf_polygons.sindex
    points = np.array([[g.x, g.y] for g in gdf_transcripts.geometry])
    
    # Get potential matches for all points at once
    possible_matches = polygons_sindex.query_bulk(points, predicate='within')
    
    # Process results
    assignments = []
    for i in range(len(gdf_transcripts)):
        matches = possible_matches[:, possible_matches[1] == i]
        if matches.shape[1] == 0:
            assignments.append('UNASSIGNED')
        elif matches.shape[1] > 1:
            assignments.append('more_than_one_matches')
        else:
            poly_id = gdf_polygons.iloc[matches[0,0]].name
            assignments.append(f"{poly_id}_polygon" if not isinstance(poly_id, str) else poly_id)
    
    return assignments
```

Final thing, very minor, `assigning_transcript` is a bit weird, how about `assign_transcripts`?

src/celldega/pre/merge_segmentations.py
```diff
@@ -0,0 +1,514 @@
...
+            new_data = {'geometry': poly}
+            new_row = gpd.GeoDataFrame([new_data])
+            new_row['cell_id'] = ID
+            new_row['technology'] = tech
+            new_row['centroid'] = poly.centroid
+            gdf_nc = gpd.GeoDataFrame(pd.concat([gdf_nc, new_row], ignore_index=True))
+
+    return gdf_nc
+
+def merge_segmentation(default_data_path, custom_data_path, output_path, clusters_within_cutout_region, inv_alpha_value_for_cutout_region=1, buffer_for_cutout_region_alpha_shape=0, ioa_small_thresh = 0.5):
```
huanlity, line 189: Consider adding some comments to show what is going on, especially between L283-L392. Most of the lines with `intersecting_cell_pairs_before_harmonization_DF` are pretty long, almost too long. No big deal but maybe you can try `black` it and see if it works?

huanlity (commented):
@jaspreetishar This is a lot of great work! My review mainly focused on the efficiency and format of the code. Two places that definitely can use some optimization are:
1. transform between micron space and pixel space:  I shared an example of what I use;
2. assign transcripts: where you can use vecterized operation instead of processing every trx one-by-one using `apply`. As we are working with many millions of points, any time saving x 200M would be huge.

I will test the notebook again once these steps are optimized.

src/celldega/pre/merge_segmentations.py
jaspreetishar: I found query_bulk required more setup and careful handling, especially with indexing. So I went with sjoin since it’s more straightforward and still performs quite well for our use case.

src/celldega/nbhd/__init__.py
```diff
@@ -4,11 +4,11 @@
 
 from libpysal.cg import alpha_shape as libpysal_alpha_shape
 import geopandas as gpd
-from shapely import Point, MultiPoint, MultiPolygon
+from shapely import Point, MultiPolygon
 from shapely.ops import transform
 import numpy as np
 import json
-from shapely.geometry import shape
+from shapely.geometry import shape, Point
```
cornhundred, line 11: it doesn't look like `Point` is being used.

src/celldega/pre/__init__.py
```diff
@@ -702,19 +728,20 @@ def add_custom_segmentation(path_landscape_files, path_segmentation_files, image
...
-                       verbose=True,
-                       segmentation_approach=segmentation_parameters['segmentation_approach'])
-
     make_meta_cell_image_coord(technology = segmentation_parameters['technology'],
                         path_transformation_matrix = os.path.join(path_landscape_files, 'micron_to_image_transform.csv'),
                         path_meta_cell_micron = os.path.join(path_segmentation_files, 'cell_metadata_micron_space.parquet'),
                         path_meta_cell_image = os.path.join(path_landscape_files, f"cell_metadata_{segmentation_parameters['segmentation_approach']}.parquet"),
                         image_scale=image_scale)
 
+    save_cbg_gene_parquets(base_path=path_landscape_files,
```
cornhundred, line 737: why is the order of these functions changed from before?

notebooks/micron_to_image_transform.csv
cornhundred, line 1: do we need this file or can it be read from the LandscapeFiles?

src/celldega/pre/merge_segmentations.py
```diff
@@ -0,0 +1,823 @@
...
+
+        max_ioa_merged = 0
+        max_ioa_merged_index = 0
+
+        if not poly.is_valid:
+            poly = make_valid(poly)
+
+        for index in possible_intersections:
+
+            poly_intersect = gdf_nc.iloc[index]["geometry"]
```
cornhundred, line 164: we can change this to use `.loc`

src/celldega/pre/merge_segmentations.py
```diff
@@ -0,0 +1,823 @@
...
+            intersecting_cell_pairs_before_harmonization_DF["cell_id_1_technology"]
+            != intersecting_cell_pairs_before_harmonization_DF["cell_id_2_technology"]
+        ]
+    )
+
+    intersecting_cell_pairs_before_harmonization_DF.reset_index(inplace=True)
+    intersecting_cell_pairs_before_harmonization_DF.drop(
+        ["index"], inplace=True, axis=1
+    )
+
```
cornhundred, line 466: Can we modularize sections of code like this into separate functions so they are easier to work with?

src/celldega/pre/merge_segmentations.py
```diff
@@ -0,0 +1,823 @@
...
+        area_1 = poly_1.area
+        area_2 = poly_2.area
+
+        area_intersection = poly_1.intersection(poly_2).area
+        area_union = poly_1.union(poly_2).area
+
+        if area_union <= 0:
+            iou = 0
+        else:
+            iou = area_intersection / area_union
```
cornhundred, line 522: can we make a function that calculates ioa and use it here to modularize?

src/celldega/pre/merge_segmentations.py
```diff
@@ -0,0 +1,823 @@
...
+            polygons_sindex=polygons_sindex,
+            gdf_polygons=gdf_polygons,
+        )
+    else:
+        gdf_transcripts["polygon_index"] = "UNASSIGNED"
+
+    return gdf_transcripts
+
+
+def assign_transcripts(gdf_polygons, gdf_transcripts):
```
cornhundred, line 69: Would we be able to reuse the code we previously used to assign transcripts to cells? Or was that work done outside the scope of Celldega in the segmentation project?

cornhundred (commented):
Hi @jaspreetishar, it is looking better but can we break down the large `merge_segmentation` method into smaller methods that are used by this user facing method? We can hide the smaller methods from the user API by starting their name with an underscore.

src/celldega/nbhd/__init__.py
jaspreetishar: it is being used in some older functions

src/celldega/pre/__init__.py
jaspreetishar: just to maintain consistency with the order in which functions are called in the run_pre_processing.py script - I can revert it if needed

notebooks/micron_to_image_transform.csv
jaspreetishar: this was probably added by mistake - I'll remove it

src/celldega/pre/merge_segmentations.py
jaspreetishar: that code was part of the segmentation project and is outside the scope of Celldega

src/celldega/nbhd/__init__.py
cornhundred: `Point` is being imported twice. Lets start using `pylint src/celldega/nbhd` to catch these.

src/celldega/pre/merge_segmentations.py
```diff
@@ -0,0 +1,749 @@
...
+            new_data = {"geometry": poly}
+            new_row = gpd.GeoDataFrame([new_data])
+            new_row["cell_id"] = ID
+            new_row["technology"] = tech
+            new_row["centroid"] = poly.centroid
+            gdf_nc = gpd.GeoDataFrame(pd.concat([gdf_nc, new_row], ignore_index=True))
+
+    return gdf_nc
+
+def boundary_cell_conflicts(largest_cutout_region_alpha_shape, cells_before_harmonization):
```
cornhundred, line 259: we need to try to simplify and add a few comments to make this easier to follow

cornhundred (commented):
Hi @jaspreetishar, some of the functions are still a little hard to follow. We can set up a time to discuss. But in short we should try to clean up the indexing logic and add some comments to make things easier to follow.

PR #67, opened 2025-02-28
[DEGA-14] break apart pre processing python code into separate files

huanlity:


huanlity:
example of running preprocessing in command line:
Screenshot 2025-02-28 at 12 55 20 PM

huanlity:
@jaspreetishar just want to share some notes with you on how to use black to format your scripts. 
1. To install, just do: `pip install black`
2. To format one, run `black -S <file_path>`, I strongly recommend using `-S` here to avoid any string normalization/line breaking.
3. You can format the whole directory by passing a dir argument after `-S` but I will just do one file by one file unless I am 100% sure I want to do it on the whole folder.

cornhundred (changes requested):
Looks good, but can we make a few changes. 
- make a method that runs all steps (like run_pre_processing.py) but that can be run in a notebook
- hide the outputs from the unzipping if possible
- skip the image downsizing step if necessary (by default we should not do this and we can skip making the file altogether)
- delete intermediate image files (e.g., prot_output_regular_downsize.tif)
- rename "xenium_transform.csv" to "micron_to_image_transform.csv" or something similar to make it tech agnostic
- maybe change make_meta_gene to take cbg rather than the path to cbg as an argument (since we have already loaded it into memory)

huanlity:
@cornhundred all changes have been implemented.

cornhundred:
Thanks I'll look over this morning.

PR #68, opened 2025-03-07
[DEGA-205] switch to custom index for cell gene names in tiled parquet files

huanlity:


src/celldega/pre/__init__.py
```diff
@@ -537,6 +541,7 @@ def save_landscape_parameters(
     tile_size=1000,
     image_info={},
     image_format='.webp',
+    use_int_index=False,
```
cornhundred, line 544: I think we should set this to True by default. What do you think?

src/celldega/pre/boundary_tile.py
```diff
@@ -272,6 +324,8 @@ def make_cell_boundary_tiles(
         path_transformation_matrix, header=None, sep=" "
     ).values
 
+    gene_str_to_int_mapping, cell_str_to_int_mapping = _get_name_mapping(path_transformation_matrix.replace('/micron_to_image_transform.csv',''))
```
cornhundred, line 327: I feel like this is a little inefficient in that its making the gene map here even though its not needed in this function. We could make an argument to tell `_get_name_mapping` which mapping we want: gene or cell.

js/vector_tile/transcripts/grab_trx_tiles_in_view.js
```diff
@@ -15,8 +15,22 @@ export const grab_trx_tiles_in_view = async (base_url, tiles_in_view, viz_state)
...
+        // When viz_state.vector_name_integer is false, extract names directly.
+        var new_trx_names_array = trx_arrow_table.getChild("name").toArray();
+      } else {
+        // When viz_state.vector_name_integer is true, map integer values to strings.
+        var new_trx_names_array = Array.from(trx_arrow_table.getChild("name").toArray()).map(num =>
+          viz_state.genes.g_nameMapping_inv[num]
+        );
+      }
+
+      console.log('trx_arrow_table.getChild("name").toArray()',trx_arrow_table.getChild("name").toArray())
```
cornhundred, line 31: drop this console log, also are we noticing changes in performance in rendering times?

js/global_variables/meta_gene.js
```diff
@@ -25,6 +25,28 @@ export const set_meta_gene = async (genes, base_url) => {
...
+        return acc;
+    }, {});
+
+    // Save the mapping and inverse mapping as cats.nameMapping_inv
+    genes.g_nameMapping = g_nameMapping;
+    genes.g_nameMapping_inv = g_nameMapping_inv;
+
     genes.gene_names = genes.gene_counts.map(gene => gene.name);
 
+    console.log('genes.g_nameMapping',genes.g_nameMapping)
```
cornhundred, line 49: drop console logs

js/global_variables/cell_names_array.js
```diff
@@ -1,5 +1,29 @@
 export const set_cell_names_array = (cats, cell_arrow_table) => {
+
+    // Extract the array of names (strings)
     cats.cell_names_array = cell_arrow_table.getChild("name").toArray();
+
+    // Create a set of unique names
+    const uniqueNames = [...new Set(cats.cell_names_array)];
```
cornhundred, line 7: I'm not sure this is necessary and will this change the order of the names and their associated indexes? I think we can just loop through the array using reduce without first getting the unique names.

cornhundred (changes requested):
Looks good, I made some comments. We'll need to spot check using an old dataset to make sure the indexes are not getting mixed up on the front-end.

src/celldega/pre/__init__.py
huanlity: It is set to`True` in the `run_pre_processing.py`. I feel like this way it is safer..

src/celldega/pre/boundary_tile.py
huanlity: fixed.

js/viz/landscape_ist.js
```diff
@@ -198,6 +198,10 @@ export const landscape_ist = async (
 
     const tmp_image_info = viz_state.img.landscape_parameters.image_info
 
+    viz_state.vector_name_integer = viz_state.img.landscape_parameters.use_int_index
+
+    console.log('viz_state.vector_name_integer', viz_state.vector_name_integer)
```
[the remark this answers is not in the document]
huanlity: fixed this and other console.log()

js/global_variables/cell_names_array.js
huanlity: Checked and everything looks good

huanlity:
I added warnings if the cell names are unique. For the trx, we don't need to do this, since we use index column directly (by design, indices won't be the same).

cornhundred:
@huanlity I forgot that we also need to use indexes for the cell-cy-gene parquets. Can you update the code that makes those and make the corresponding front end changes? This will improve gene expression overlaying performance.

huanlity:
@cornhundred , I have tested and uploaded the updated `cbg` folder to this [repo](https://github.com/broadinstitute/celldega_Xenium_V1_human_Pancreas_FFPE_outs_smaller) for other people to validate ( remember to pull latest from branch DEGA-205…). It works on my end. On the Pancreas dataset, the total size of cbg files was reduced by **40%** after using integer indices.

```
base_url = 'https://raw.githubusercontent.com/broadinstitute/celldega_Xenium_V1_human_Pancreas_FFPE_outs_smaller/main/Xenium_V1_human_Pancreas_FFPE_outs'

landscape_ist = dega.viz.Landscape(
    technology='Xenium',
    base_url = base_url,
)

landscape_ist
```

js/global_variables/cell_exp_array.js
```diff
@@ -16,28 +16,22 @@ export const update_cell_exp_array = async (cats, genes, base_url, inst_gene, ve
...
             // Apply logarithmic transformation
             const log_exp_value = Math.log1p(exp_value); // log1p(x) = log(1 + x)
 
             // Scale to 0-255 range using the log of the max expression value
             const log_max_exp = Math.log1p(max_exp); // log1p(x) = log(1 + x)
-            new_exp_array[index] = (log_exp_value / log_max_exp) * 255;
+            new_exp_array[name] = (log_exp_value / log_max_exp) * 255;
         } else {
-            console.log('Cell name not found in cell_name_to_index_map');
+            console.log('Cell name not found in cats.nameMapping_inv');
```
cornhundred, line 32: lets remove this console log since it slows down the browser if there is an error.

cornhundred (changes requested):
It looks good, but I am not able to get older datasets that were uploaded to GitHub that do not use an index or their LandscapeFiles to correctly color cells based on their gene expression. 

```
base_url = 'https://raw.githubusercontent.com/broadinstitute/celldega_Xenium_Prime_Mouse_Brain_Coronal_FF_outs/main/Xenium_Prime_Mouse_Brain_Coronal_FF_outs'

landscape_ist = dega.viz.Landscape(
    technology='Xenium',
    base_url = base_url,
)

landscape_ist
```

I'm seeing the console log "cell name not found in cats.nameMapping_inv" (which we should remove so as to not crash the browser if there is a mixup in the indexing - we can think of a way to have a single console log happen rather than one for each cell).

cornhundred:
Thanks @huanlity, this worked on the Pancreas dataset and ran on the older GitHub hosted LandscapeFiles.

PR #72, opened 2025-04-03
working on aws s3 reading

cornhundred:
This pull request adds the ability to make GET requests from private AWS buckets (e.g., on Manifold.ai). We are using aws4fetch to do this. I am currently getting aws4fetch from a CDN rather than updating the package.json because we are running into breaking changes with the new release of luma.gl/deck.gl version 9.1.7 that we will avoid for now. 

Current issues are
* image layers are not rendering - likely because of custom shader changes
* community editable layer not working - possibly some kind of api change

Note: @huanlity and @jaspreetishar  **do not run** `npm install` because this will upgrade dependencies. If you have to re-install, then run `npm ci` for a clean install that uses the unchanged `package-lock.json`.


js/viz/landscape_ist.js
```diff
@@ -50,13 +52,15 @@ export const landscape_ist = async (
     umap={},
     landscape_state='spatial',
     segmentation='default',
+    creds={},
     view_change_custom_callback=null
 ) => {
 
+    console.log('2025-04-08 - updates')
```
huanlity, line 59: Do we still need this?

js/viz/landscape_ist.js
```diff
@@ -167,8 +213,8 @@ export const landscape_ist = async (
         viz_state.umap.state = true
     }
 
-    console.log(landscape_state)
-    console.log('viz_state.umap.state', viz_state.umap.state)
+    // console.log(landscape_state)
+    // console.log('viz_state.umap.state', viz_state.umap.state)
```
huanlity, line 217: please consider removing.

huanlity (commented):
There are many changed files but they are pretty much focused on adding one argument: `viz_state.aws`. All looks to me except the current issue you mentioned in the  PR.

PR #76, opened 2025-04-22
Dega 110 hexagon tile geo data frame

jaspreetishar:


src/celldega/nbhd/test_hextile.py
```diff
@@ -0,0 +1,14 @@
+from . import *
+from ..pre.merge_segmentations import *
+
+def test_hextile():
+
+    gdf_dummy_trx = generate_random_points_gdf(n_points=50, x_range=(0, 100), y_range=(0, 100))
+    dummy_img = generate_dummy_image(width=100, height=100)
```
cornhundred, line 7: we can drop the generate_dummy_image method since we are just making an image in order to calculate the area of the image and just calculate the intended area instead.

src/celldega/nbhd/__init__.py
```diff
@@ -205,3 +213,301 @@ def alpha_shape_geojson(gdf_alpha, meta_cluster, inst_alpha):
...
+
+
+def hextile_area(gdf_hextile, tissue_img_shape):
+
+    return round(sum(gdf_hextile.area.tolist())) == round(
+        tissue_img_shape[0] * tissue_img_shape[1]
+    )
+
+
+def hextile_trx_assignment(hextile_assigned_trx):
```
cornhundred, line 302: can we rename this method to indicate that this is a check on the consistency rather than a assignment calculation

notebooks/hexatile_view_of_unassigned_trx.ipynb
cornhundred, line 1: Can we drop this notebook from this pull request in order to stay in scope where we just need to make a hexagonal tiling neighborhood? We can make a new notebook that just demonstrates this - make a hexagon neighborhood collection as a geodataframe.

cornhundred (changes requested):
Can we limit the scope of this pull request to just make the hexagonal neighborhood collection and move the notebooks and code that calculate the transcript assignment assignment (which look great) to a pull request for DEGA-17?

src/celldega/nbhd/__init__.py
```diff
@@ -205,3 +213,301 @@ def alpha_shape_geojson(gdf_alpha, meta_cluster, inst_alpha):
     geojson_alpha['inst_alpha'] = inst_alpha
 
     return geojson_alpha
+
+def create_hextile(radius, path_landscape_files=None, img_height=100, img_width=100):
```
huanlity, line 213: I don't think you need to change it now, but just a heads-up that we may change this later so we calculate the hex bins in micron space, as last time I met @cornhundred , he wanted everything (ALPH. HEX, GRAD) in micron space and only transform it to pixel space right before front-end rendering. I will comment a code snippet I tested on your Jira page.

src/celldega/nbhd/__init__.py
```diff
@@ -205,3 +209,81 @@ def alpha_shape_geojson(gdf_alpha, meta_cluster, inst_alpha):
...
+        ax.set_xlabel("x (pixels)", fontsize=25)
+        ax.set_ylabel("y (pixels)", fontsize=25)
+        plt.xticks(fontsize=20)
+        plt.yticks(fontsize=20)
+        plt.show()
+        plt.close()
+
+    return gdf_hextile
+
+def hextile_area(gdf_hextile, img_width, img_height):
```
huanlity, line 285: It is only meaningful if the area is calculated in physical space and not pixel space.

jaspreetishar: Thank you for pointing that out. the area is being calculated in pixel space for a pytest to ensure correct hextile generation across the full tissue

cornhundred: I think we can remove the `hextile_area` function and move the logit into the test_hextile.py file.

src/celldega/nbhd/test_hextile.py
```diff
@@ -0,0 +1,7 @@
+from . import *
+
+def test_hextile():
+
+    gdf_dummy_hextile = create_hextile(radius=5, img_width=100, img_height=100)
+
+    assert hextile_area(gdf_dummy_hextile, img_width=100, img_height=100) is True
```
cornhundred, line 7: can we move this logic 

```
round(sum(gdf_hextile.area.tolist())) == round(
        img_width * img_height
``` 

into the assert statement instead of requiring a function?

src/celldega/nbhd/__init__.py
```diff
@@ -205,3 +211,108 @@ def alpha_shape_geojson(gdf_alpha, meta_cluster, inst_alpha):
     geojson_alpha['inst_alpha'] = inst_alpha
 
     return geojson_alpha
+
+def create_hextile(radius, path_landscape_files=None, img_height=100, img_width=100, pixel_size=0.2125):
+
+    if isinstance(path_landscape_files, str):
```
cornhundred, line 217: This is ok for now, but we should add the image height/width to the `landscape_parameters.json` in a future pull request (DEGA-244).

cornhundred (changes requested):
Hi @jaspreetishar, I requested a change to the test (remove function and move logic into test) but otherwise looks good.

src/celldega/nbhd/test_hextile.py
jaspreetishar: Moved to assert, looks much cleaner now

cornhundred (changes requested):
Sorry for the late request, but can we try to generate a single hexagon and then translate it repeatedly to generate the GeoDataFrame?

ChatGPT suggests something like this 

```
# Precompute unit hexagon
angles = np.radians(np.arange(0, 360, 60))
unit_hex = Polygon([(radius * np.sin(a), radius * np.cos(a)) for a in angles])

# Generate hexagons by translating the unit hex
hexagons = []
for row in range(n_rows):
    for col in range(n_cols):
        x = col * horiz_spacing
        y = row * vert_spacing
        if row % 2 == 1:
            x += horiz_spacing / 2
        hexagons.append(translate(unit_hex, xoff=x, yoff=y))

```

cornhundred (approved):

PR #149, opened 2025-07-15
Pass auth token to image tile loader (#148)

cornhundred:
* fix image tile loader to include auth headers

* making alpha release to test Terra auth fix

* re-built widget.js


js/deck-gl/utils/tiles.js
```diff
@@ -28,14 +28,21 @@ export const create_get_tile_data = (
...
+    const { fetch: fetchOptions = {}, ...loaderOptions } = options;
+
+    // Fetch function that preserves headers and supports aws signed requests
+    const fetch_fn = async (url, fetchOpts = {}) => {
+      const merged = { ...fetchOptions, ...fetchOpts };
+      if (aws) {
+        return aws.fetch(url, merged);
+      }
+      return fetch(url, merged);
+    };
```
huanlity, line 41: Does this fix `fetch` error when visualizing Visium-HD data in Celldega?

cornhundred: No, this doesn't fix that. 

I'm thinking we will want to take the approach of visualizing Visium HD data using landscape_ist for now since the newer versions have segmented cells/transcripts. But we can discuss this in another pull request.

PR #196, opened 2025-08-04
Add hex tile clustering utilities (#185)

cornhundred:
* Add hex tile clustering

* initial set up

* Add niche clustering example notebook

* adding gdf_niche/gdf_hex clustergram visual

* adding niche labels as meta_row for clustergram

* replacing percentages with proportions in calc_nbp

* fixing function name

* fixing return for prepare_nbhd_for_clustering

* rearranging notebook

* making prepare_nbhd_for_clustering inclusive of other nbhd types

* using inclusive gdf names

* clustergram transposed

* visualizing in landscape viewer: initial setup

* moving imports to the start


src/celldega/nbhd/hextile.py
```diff
@@ -38,9 +42,51 @@ def generate_hex_grid(
...
-        {
-            "name": [f"hex_{i}" for i in range(len(hexagons))],
-            "geometry": hexagons,
-        },
+        {"name": [f"hex_{i}" for i in range(len(hexagons))], "geometry": hexagons},
         crs=gdf_cell.crs,
     )
+
+
+def prepare_nbhd_for_clustering(
```
cornhundred, line 50: I think we can drop this function in favor of improving the API for calc_nbp

src/celldega/nbhd/hextile.py
```diff
@@ -38,9 +42,51 @@ def generate_hex_grid(
...
+
+    elif nbhd_type == 'GRAD':
+        gdf_object = _get_gdf_cell(adata)
+        gdf_micron = gpd.read_parquet(f'{path_landscape_files}/spatial_regions/sketched_regions_micron.parquet')
+        gdf_roi = gdf_micron.loc[gdf_micron['roi'] == 'region_1']
+        gdf_nbhd = calc_grad_nbhd_from_roi(gdf_roi, gdf_object, 200)
+        gdf_nbhd['name'] = gdf_nbhd['band']
+
+    # Calculate cell-type counts per nbhd tile
+    counts, population_distribution = calc_nbp(gdf_object, gdf_nbhd)
```
cornhundred, line 83: We need to add an argument for the cell category that we want to use to calculate the cell distributions (e.g., leiden, cell type, etc). Also, we can add an additional optional argument adata (or replace the existing gdf_object - probably keep it in case a user already has a gdf) - only allow a user to have adata or gdf - another option is to have an argument called data and check the type. This is what we do with the Matrix class in the clust method.

I think our workflow should be that the user 
* makes a hextile neighborhood collection at their desired resolution
* user passes this in with the AnnData and their category of interest to obtain a nbp matrix (which is returned as ad_tiles - lets call this `adata_nbp` for clarity).

src/celldega/nbhd/hextile.py
```diff
@@ -38,9 +42,51 @@ def generate_hex_grid(
...
+
+    # Prepare AnnData for clustering
+    ad_tiles = AnnData(
+        X=population_distribution.values,
+        obs=pd.DataFrame(index=population_distribution.index),
+        var=pd.DataFrame(index=population_distribution.columns),
+    )
+
+    return ad_tiles, gdf_nbhd
```
cornhundred, line 92: lets get rid of this function and this will simplify the outputs so that we do not have to pass back the gdf_nbhd to the user since this will be an argument instead.

cornhundred (commented):
Requested API changes

cornhundred:
* Lets make a method for melting

cornhundred:
Hi @jaspreetishar, can you 
* update the niche_clustering notebook to: 
    * remove the synthetic data section
    * use the new calc_nbp API
* change the API for generate_hex_grid to accept diameter instead of radius (I feel this is slightly easier)
* re-run the hextile neighborhood analysis using a diameter of 50um
* filter out hexagon tiles that have fewer than 5 cells (or some threshold that makes sense)

src/celldega/nbhd/hextile.py
jaspreetishar: fixed

PR #200, opened 2025-08-05
Dega 308 umap

cornhundred:


cornhundred:
This pull request also fixes a bug with the clust method's Matrix class in the downsample method.

PR #230, opened 2025-10-03
Dega 382 visium hd gallery example

jaspreetishar:
Created a gallery example for Visium HD Jitter Mouse Brain dataset


cornhundred (changes requested):
Looks great, but could use the term "cell segmentation" instaead of "jitter" if you don't mind changing some of the file names etc? This is the real improvement compared to the previous visualization. Otherwise, looks good to merge. I can take care of the version release and documentation update.

PR #252, opened 2025-11-14
Add configurable rotation support to landscape view

cornhundred:
## Summary
- add a rotate trait to the Landscape widget API and forward it through the front-end renderer
- introduce shared rotation utilities and apply the rotation matrix to all deck.gl landscape layers
- update viewport calculations to fetch tiles and filter data correctly when the visualization is rotated

## Testing
- pytest --override-ini addopts="" tests/unit/test_viz/test_widget.py

------
[Codex Task](https://chatgpt.com/codex/tasks/task_b_6916a5d313c8832a89c51a7438a15479)

### DIFF
--- a/js/deck-gl/layers/trx_layer.js
+++ b/js/deck-gl/layers/trx_layer.js
@@ -61,7 +61,7 @@ export const ini_trx_layer = (viz_state) => {
         genes.selected_genes.length === 0 ||
         genes.selected_genes.includes(inst_gene)
           ? 255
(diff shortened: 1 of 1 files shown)


cornhundred:
I used Codex to implement a rotation argument for the Landscape widget. It works in my testing and the implementation looks ok. I'm gong to use it to rotate a Xenium dataset to match the orientation with a Visium HD dataset.

PR #287, opened 2025-12-20
add notebook for xenium protein T cell work

huanlity:


cornhundred:
This will be implemented in a different pull request using an updated neighborhood API.

PR #293, opened 2026-03-03
improve gradient function

huanlity:
The previous function `calc_grad_nbhd_from_roi()` has these limitations and therefore is depreciated as `_calc_grad_nbhd_from_roi()`. Once this PR is reviewed and approved, we can probably remove it:
1. Only work with gdf in micron space
2. Only go inside and not outside
3. Missing `name`,`color`,`area` that are handy to use Landscape to visualize rings
4. Missing flexibility to set max distance

A new function is created: `gradient_from_gdf()`. Improvemnts:1
1. Work with gdf in micron space like manual drawing as well as gdf in pixel space like the alpha shape created using Celldega
2. Easy to set bin width and max distance, and goes both inward and outward
3. Populate `name`,`color`,`area` so the gradient nbhd can be visualized in Landscape properly

Example:

"""
scale_um_per_pixel=dega.nbhd._get_micron_per_pixel(technology)
bin_width=10
max_dist_um=50

gdf_rings_merged, name_order = dega.nbhd.gradient_from_gdf(
    gdf_alpha_tumor,
    scale_um_per_pixel,
    bin_width,
    max_dist_um,
    is_pixel_space=True,
)
"""

### DIFF
--- a/src/celldega/nbhd/__init__.py
+++ b/src/celldega/nbhd/__init__.py
@@ -1,7 +1,7 @@
 """Module for performing neighborhood analysis."""
 
 from .alpha_shapes import alpha_shape, alpha_shape_cell_clusters, filter_alpha_shapes
-from .gradient import calc_grad_nbhd_from_roi
+from .gradient import gradient_from_gdf
 from .hextile import generate_hextile, hextile_niche
 from .neighborhoods import (
     NBHD,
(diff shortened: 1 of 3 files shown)


cornhundred:
This was implemented in a different pull request.

PR #294, opened 2026-03-12
Fix dim Xenium morphology tiles and improve OME-TIFF loading

jaspreetishar:
Fix dim Xenium morphology tiles and improve OME-TIFF handling

1. Newer Xenium morphology images have lower raw intensity values (~10× lower than older datasets), which causes tiles to appear dim when displayed directly. 
- Apply per-channel intensity windowing using a high-percentile upper bound (default: 99)
   - `upper_percentile` argument is now added to `dega.pre.main()`
- Clip extreme bright pixels before display scaling
- Normalize each channel independently for more stable visualization
- Add optional gamma support for channel tone adjustment (default: 1)
- Control channel brightness with a capped 8-bit white level (default: 40)
   - Instead of stretching the normalized signal to the full 0–255 display range, we cap it at 40. This intentionally compresses brightness, preventing channels from appearing blown out.
- Keep non-DAPI scaling neutral by default to avoid over-bright composites

3. Read Xenium OME-TIFF channels individually using tifffile to avoid loading the full image stack into memory.

4. Detect morphology images using the `*0000*.ome.tif` pattern to support newer Xenium image naming.

** successfully tested the code changes on both private and public Xenium data.


src/celldega/pre/trx_tile.py
```diff
@@ -627,7 +627,7 @@ def _collect_tile_data_for_row_groups(
     tile_data_list = []
     for tile_i in tqdm(range(n_tiles_x), desc="Ordering tiles"):
         for tile_j in range(n_tiles_y):
-            tile_df = tile_dict.get((tile_i, tile_j), None)
```
jaspreetishar: ruff is flagging SIM910, which means passing None explicitly as the default to dict.get is unnecessary because None is already the default.

src/celldega/pre/boundary_tile.py
```diff
@@ -493,7 +493,7 @@ def _collect_boundary_tile_data_for_row_groups(
     tile_data_list = []
     for tile_i in tqdm(range(n_tiles_x), desc="Ordering tiles"):
         for tile_j in range(n_tiles_y):
-            tile_df = tile_dict.get((tile_i, tile_j), None)
```
jaspreetishar: ruff is flagging SIM910, which means passing None explicitly as the default to dict.get is unnecessary because None is already the default.

src/celldega/pre/__init__.py
```diff
@@ -458,40 +489,71 @@ def remove_intermediate_files(path_landscape_files):
...
+        for channel_info in channel_map:
+            channel_name = channel_info["name"]
+            channel_index = channel_info["index"]
+
+            print(f"Reading channel '{channel_name}' (index {channel_index})")
+
+            # Read one channel at a time to avoid large memory usage
+            channel_2d = series.asarray(key=channel_index)
+
+            _process_image_channel(
```
huanlity, line 563: What is the reason to leave the `white-level` argument out here? Does its default value (40) work for other datasets? Thanks!

src/celldega/pre/__init__.py
```diff
@@ -458,40 +489,71 @@ def remove_intermediate_files(path_landscape_files):
...
+            )
+
+        for channel_info in channel_map:
+            channel_name = channel_info["name"]
+            channel_index = channel_info["index"]
+
+            print(f"Reading channel '{channel_name}' (index {channel_index})")
+
+            # Read one channel at a time to avoid large memory usage
+            channel_2d = series.asarray(key=channel_index)
```
huanlity, line 561: I like this part

src/celldega/pre/boundary_tile.py
huanlity: Thanks. Did you test if removing `, None` break the pipeline or generate unexpected results? If not, we can probably fix it with ruff. @cornhundred any thought?

src/celldega/pre/__init__.py
jaspreetishar: the default value of 40 worked for both a private dataset I used to test the new code changes and the publicly available Xenium pancreas dataset. that said, there may be cases where this parameter needs to be adjusted. I can add it to dega.pre.main(), similar to how I added upper_percentile, if you think it would be useful

huanlity: We either use an argument or not use one but we cannot have an argument and not use it. A user should have a way to adjust `white-level` without going deep to our helper function `_process_image_channel()` IMO.

src/celldega/pre/__init__.py
```diff
@@ -331,43 +331,67 @@ def create_cluster_and_meta_cluster(
...
-    # Extract and process the channel
-    scale = 1 if channel_name.lower() == "dapi" else 2  # Adjust intensity for better visualization
-    if img.ndim == 3:
-        image_data = img[..., channel_index] * scale
-    elif img.ndim == 2:
-        image_data = img * scale
+    if img.ndim != 2:
+        raise ValueError(f"Expected a 2D channel image, got shape {img.shape}")
+
+    channel = img.astype(np.float32)
```
huanlity, line 364: Do we have to have this line?
`channel = img.astype(np.float32)`

src/celldega/pre/__init__.py
```diff
@@ -331,43 +331,67 @@ def create_cluster_and_meta_cluster(
...
+
+    # Keep scaling neutral unless you intentionally want a boost
+    scale = 1.0 if channel_name.lower() == "dapi" else scale_non_dapi
+    channel *= scale
+
+    # Per-channel display window
+    lo = float(channel.min())
+    hi = float(np.percentile(channel, upper_percentile))
```
huanlity: (Gemini suggested)

Potential Problem: np.percentile is computationally expensive because it involves sorting the data. On a large image (e.g., 20k x 20k pixels), this line will take significantly longer than the rest of the function combined.

Improvement: Use subsampling. Calculating the percentile on a 1/10th scale version of the image (e.g., channel[::10, ::10]) will give you an almost identical hi value in a fraction of the time.

huanlity (commented):
Thanks for working on this and making the ome-tif reading part more efficiently, these are super helpful. Please see my individual comments above and let me know if you have questions.

src/celldega/pre/__init__.py
jaspreetishar: it converts the image to floating point so we can safely do normalization, scaling, and gamma correction without losing precision

src/celldega/pre/boundary_tile.py
jaspreetishar: I tested use_row_groups=True without the `None` fallback in tile_dict.get((tile_i, tile_j)), and the results are as expected.

src/celldega/pre/__init__.py
jaspreetishar: makes sense. I have added `white_level` as another parameter (default=40) in dega.pre.main()

src/celldega/pre/__init__.py
jaspreetishar: thank you for the suggestion - I have implemented subsampling

jaspreetishar:
Thank you for the helpful review, @huanlity! I have addressed your questions and implemented your suggestions. Please let me know your thoughts.

src/celldega/pre/__init__.py
huanlity: Good, do you see any run time difference before and after this implementation? Hopefully it is not huge.

huanlity (approved):
I tested the preprocess notebook and everything worked as expected. Great work! I approved the PR but maybe we can wait for Nick to take a look when he is back before merging.

src/celldega/pre/__init__.py
jaspreetishar: `channel = img.astype(np.float32)` is not memory-intensive, so it runs quickly. Were you by any chance referring to this comment instead, https://github.com/broadinstitute/celldega/pull/294#discussion_r2942885988 ?

jaspreetishar:
> I tested the preprocess notebook and everything worked as expected. Great work! I approved the PR but maybe we can wait for Nick to take a look when he is back before merging.

Yes, that sounds good to me - thank you for your review!

src/celldega/pre/__init__.py
huanlity: No I was referring to `np.float32`, good to hear it runs quickly. But yeah I am curious now if the subsampling below helps with the speed?

jaspreetishar: I had not formally timed the two approaches before, but I just did using the private dataset. The older code took 10.70 minutes, while the version you suggested completed in 9.36 minutes.

src/celldega/pre/run_pre_processing.py
```diff
@@ -155,6 +155,8 @@ def main(
     max_workers=1,
     use_row_groups=False,
     max_row_groups_per_file=400,
+    upper_percentile=99,
+    white_level=40,
 ):
```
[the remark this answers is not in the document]
jaspreetishar: applied docstring-related suggestion only

tests/unit/test_pre/test_chromium.py
```diff
@@ -6,6 +6,13 @@
 import pandas as pd
 import pytest
 
+import sys
+import types
+
+fake_module = types.ModuleType("celldega.pre.run_pre_processing")
+fake_module.main = lambda *args, **kwargs: None
+sys.modules["celldega.pre.run_pre_processing"] = fake_module
+
```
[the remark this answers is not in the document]
jaspreetishar: ignored

tests/unit/test_pre/test_row_groups.py
```diff
@@ -238,11 +255,11 @@ def test_landscape_parameters_structure(self, tmp_path):
         }
 
         output_path = tmp_path / "landscape_parameters.json"
-        with open(output_path, "w") as f:
+        with Path.open(output_path, "w") as f:
             json.dump(params, f, indent=2)
 
         # Read back and verify
-        with open(output_path) as f:
+        with Path.open(output_path) as f:
```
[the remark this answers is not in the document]
jaspreetishar: suggestion ignored as ruff expects Path()

src/celldega/pre/__init__.py
```diff
@@ -331,43 +331,69 @@ def create_cluster_and_meta_cluster(
...
+    channel = img.astype(np.float32)
+
+    # Keep scaling neutral unless you intentionally want a boost
+    scale = 1.0 if channel_name.lower() == "dapi" else scale_non_dapi
+    channel *= scale
+
+    # Per-channel display window
+    lo = float(channel.min())
+
+    sample = channel[::10, ::10]
```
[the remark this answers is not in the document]
jaspreetishar: accepted the suggestion

src/celldega/pre/__init__.py
```diff
@@ -331,43 +331,69 @@ def create_cluster_and_meta_cluster(
...
+        channel = np.clip(channel, lo, hi)
+
+        # Normalize
+        channel = (channel - lo) / (hi - lo)
+
+        # Optional gamma correction (gamma < 1 brightens midtones)
+        if gamma != 1.0:
+            channel = np.power(channel, gamma)
+
+        image_data = (channel * white_level).astype(np.uint8)
```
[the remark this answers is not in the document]
jaspreetishar: accepted the suggestion

src/celldega/pre/__init__.py
```diff
@@ -376,7 +402,14 @@ def _process_image_channel(path_landscape_files, channel_info, img):
...
 
-def create_image_tiles(technology, data_dir, path_landscape_files, image_tile_layer="dapi"):
+def create_image_tiles(
+    technology,
+    data_dir,
+    path_landscape_files,
+    image_tile_layer="dapi",
+    upper_percentile=99,
+    white_level=40,
+):
```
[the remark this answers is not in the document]
jaspreetishar: accepted the suggestion

cornhundred:
## Image intensity normalization for Xenium tiles

Reworks Xenium morphology image tiling to use **per-channel intensity windowing**
(à la Xenium Explorer) instead of a fixed brightness multiplier.

### Changes
- `_process_image_channel` now clips each channel at a configurable `upper_percentile`,
  normalizes, and maps to an 8-bit display range scaled by `white_level`.
- New tunable kwargs (backward-compatible, sensible defaults):
  - `create_image_tiles(..., upper_percentile=99, white_level=100)`
  - `create_image_tiles_xenium(..., upper_percentile=99, white_level=100)`
  - `run_pre_processing.main(..., upper_percentile=99, white_level=100)`
- Reads one channel at a time from the OME-TIFF (`series.asarray(key=...)`) to avoid
  loading the full multi-channel image into memory.
- Minor: `dict.get(key, None)` → `dict.get(key)` in `boundary_tile`/`trx_tile`.

### Merge note (main → branch)
Resolved conflicts from the DEGA-487 cluster-collection merge by combining both sides:
- **Kept** this branch's intensity logic.
- **Adopted** main's `resolve_xenium_morphology_ome_path()` for locating the morphology
  OME-TIFF (handles newer/Atera filenames + `morphology.ome.tif` fallback), replacing the
  hardcoded `*0000*.ome.tif` glob.
- **Adopted** main's resolver-based `_check_required_files`.
- Renamed the internal `path_landscape_files` param to `path_dega_files` for consistency
  (the old name was undefined in `run_pre_processing.main`).

### ⚠️ Behavior change
Non-DAPI channels no longer receive the previous `2×` scale boost; brightness is now
controlled by `upper_percentile`/`white_level`. Regenerated tiles will look different.

### Testing
- `ruff check` / `ruff format` clean
- `pytest tests/unit/test_pre/` — all passing

PR #296, opened 2026-04-11
Bikes

cornhundred:


cornhundred:
## Browser-driven matrix axis slicing

Makes the **browser the source of truth for `net_mat`**, so matrix slices can be
pulled directly from the front-end — including in standalone exported HTML where
no Python kernel is running.

### What changed

- **New traits on the Matrix/Clustergram widget** (`src/celldega/viz/widget.py`):
  - `matrix_slice_request` (:1016) — set from Python (or another front-end) to
    request a slice: `{req_id, op, ...}` for `row`/`col` (by `index`), `cell` (by
    `row`/`col`), or `row_col` (by `row_index`/`col_index`, optional `max_entries`).
  - `matrix_slice_result` (:1019) — populated by the Matrix front-end in response,
    with `slice_kind`, `entries`, `matrix_convention`, etc.
  - `request_matrix_slice(...)` (:1305) — blocking Python helper that emits a
    request and polls for the matching `req_id` (default `timeout=5.0s`). Or link
    another widget without a Python round-trip:
    `jslink((cgm, "matrix_slice_result"), ...)`.

- **Front-end slicing** (`js/matrix/matrix_axis_slice.js`, new — pure functions,
  no test coverage yet): `buildRowAxisSlice`/`buildColAxisSlice` scan `net_mat`,
  drop zero/non-finite entries, sort descending by value, cap at `max_entries`
  (default 2000; `< 0` = all, capped at `UNBOUNDED_AXIS_CAP` = 500,000).
  `js/viz/matrix_viz.js:334-384` wires `change:matrix_slice_request` to these
  builders and writes the result back via `matrix_slice_result`.

- **Auto-request on click** (`js/deck-gl/matrix/label_layers.js`,
  `js/deck-gl/matrix/mat_layer.js`): row/col label clicks and cell clicks now
  call `emitMatrixSliceRequest` directly (row/col use `max_entries: -1`, i.e.
  request *all* non-zero entries) alongside the existing `click_info` update.
  **Behavior change**: every matrix click now triggers a full axis-slice
  computation client-side, not just explicit Python-initiated requests.

- **Cache fix** (`src/celldega/clust/matrix.py:674-680`): rebuild `node_info` from
  current `row_attr`/`col_attr` before export, so exported `row_nodes`/`col_nodes`
  no longer miss `cat-*` keys after mutating attr lists (e.g. appending a column
  after `add_category`).

### How to test manually

Python side (needs a live kernel):
```python
cgm.request_matrix_slice("row", index=0, max_entries=20)
cgm.request_matrix_slice("cell", row=0, col=0)
```
Front-end only (works in exported standalone HTML too): click a row/column label
or a matrix cell and inspect `matrix_slice_result` via `jslink` or the browser
console (`model.get("matrix_slice_result")`).

### Review focus — please double-check before merge

1. **The cache fix forces a full rebuild on every `make_viz()` call**
   (`matrix.py:674-680`): `self._dat_cache = None` / `_dirty_flags[DATA] = True`
   now run unconditionally at the top of `make_viz()`, even though `clust()`,
   `add_category`, and `add_cats` already call
   `self._invalidate_cache(CacheLevel.DATA.value)` before invoking it. This
   defeats the lazy-cache design for every `make_viz()` call, not just the one
   scenario it's meant to fix (direct `row_attr`/`col_attr` mutation, which has
   no invalidation hook). Consider invalidating at the actual mutation point
   instead of blanket-invalidating inside `make_viz()`, or at minimum call the
   existing `self._invalidate_cache(CacheLevel.DATA.value)` helper there instead
   of duplicating its logic by hand (that helper also evicts `_ranking_cache`,
   which the manual version currently skips).
2. **No JS test coverage** for the new pure functions in `matrix_axis_slice.js`
   (`buildRowAxisSlice`, `buildColAxisSlice`, `buildCellSlice`,
   `buildRowColPairSlice`) — straightforward to unit test, and `js/__tests__/`
   already has a pattern for similar pure modules (`cell_color.test.js`,
   `table_accessors.test.js`).
3. No `CHANGELOG.md` entry yet for this feature.

### Notes

- Merged latest `main` (current version `0.18.1`).
- `docs/assets/js/celldega.js` bundle regenerated via `node build.js`.
- `ruff format --check` and `prettier --check` both clean on the changed files
  (verified).
- `request_matrix_slice`'s blocking wait was tested manually and confirmed
  working end-to-end (does not time out).

PR #298, opened 2026-04-22
Atera branch

cornhundred:


cornhundred:
ChatGPT Summary of pull request 

**Per-File Overview**
Frontend:
- celldega.js: rebuilt/minified docs bundle that mirrors the source JS changes.
- calc_viewport.js: moves close-up bar aggregation to cached typed-array counting, adds viewport caches, and avoids reloading trx/path data when visible tiles have not changed.
- cell_layer.js: stores cells in compact category/position buffers instead of object arrays for downstream aggregation.
- trx_layer.js: switches transcript coloring/picking from gene-name arrays to integer gene IDs plus indexed color lookup.
- tooltips.js: transcript tooltips now resolve names from `trx_gene_ids` and `g_nameMapping_inv`.
- color_dict_gene.js: builds an integer-indexed color lookup (`g_colorMapping_inv`) alongside the existing name-keyed map.
- meta_gene.js: adds forward gene-name-to-ID mapping (`g_nameMapping`) to complement the reverse lookup.
- selected_genes.js: centralizes maintenance of `selected_gene_ids` and makes sync/update paths more defensive.
- cbg_row_group_reader.js: adds fallback discovery for chunked `cbg/chunk_0.parquet` datasets when a legacy single-file CBG path is missing.
- get_scatter_data.js: replaces repeated typed-array concatenation with a two-pass single-allocation flatten.
- image_row_group_reader.js: can recover `zoom_info` from parquet metadata when landscape parameters omit it.
- row_group_tile_reader.js: factors row-group reads into a shared helper, adds a small request cache, parallelizes per-file chunk reads, and can return either merged tables or table arrays.
- switch_dataset.j: rebuilds compact cell state during dataset switches and resets transcript/viewport caches while preserving persistent selections.
- ui_containers.js: prevents bar containers from auto-scrolling back to top while in close-up mode.
- yearbook_ui.js: same close-up scroll preservation for Yearbook bar panels.
- compact_data.js: new helper module for empty compact structures, visible-tile keys, bar equality checks, and compact cell encoding.
- grab_trx_tiles_in_view.js: materializes transcript positions and gene IDs into compact typed buffers from either row-group or per-tile parquet reads.
- landscape_ist.js: initializes compact transcript/cell state, viewport caches, and `selected_gene_ids`.
- landscape_sst.js: initializes `selected_gene_ids` and transcript gene-ID storage for SST mode too.
- yearbook.js: ports Yearbook gene/cell bar aggregation to compact buffers and scratch-count arrays, and initializes the same compact state used in Landscape.
- update_cell_clusters.js: rebuilds compact cell category IDs after reclustering and invalidates cached cell-bar data.

Python/preprocessing:
- Atera_viz.ipynb: new example notebook that loads a remote Atera Xenium-style dataset, reads `df_sig.parquet`, and links a landscape with a clustergram.
- Xenium_pre-process.ipynb: notebook logic is effectively unchanged; the diff mainly strips widget state and updates metadata.
- neighborhoods.py: NBHD image-based path now resolves Xenium morphology TIFFs through the new helper instead of hard-coding one filename.
- __init__.py]: uses the new morphology resolver in preprocessing, relaxes required-file checks for newer Xenium/Atera layouts, and adds a clearer error for `image_tile_layer='all'` when only `morphology.ome.tif` exists.
- image_info.py: new `resolve_xenium_morphology_ome_path()` helper that supports classic Xenium and alternate Atera-style morphology file layouts.
- trx_tile.py: big preprocessing change that adds disk-backed streaming tile assignment for large transcript datasets, with shard/spill helpers for both legacy tile files and row-group parquet output.

Artifacts:
- package-lock.json: refreshes the lockfile to the current package version and updates a lot of peer metadata; it looks like lockfile churn rather than a meaningful dependency change.

**Checks**
`npm run test:js:fast` passed. `npm run lint:js` passed with 8 `prefer-destructuring` warnings in [calc_viewport.js](/Users/feni/Documents/celldega/js/deck-gl/core/calc_viewport.js:64) and [yearbook.js](/Users/feni/Documents/celldega/js/viz/yearbook.js:578), but no lint errors.

If you want, I can turn this into a PR-ready summary comment or a tighter GitHub review with just the actionable findings.

Frontend:

The branch mainly speeds up transcript-heavy close-up views. The biggest change is that transcript and cell data stopped being handled as lots of little JS objects and started being handled as compact typed-array buffers with integer IDs in [grab_trx_tiles_in_view.js](/Users/feni/Documents/celldega/js/vector_tile/transcripts/grab_trx_tiles_in_view.js), [compact_data.js](/Users/feni/Documents/celldega/js/utils/compact_data.js), and [calc_viewport.js](/Users/feni/Documents/celldega/js/deck-gl/core/calc_viewport.js). It also avoids redoing expensive viewport work unless the visible tile set actually changed, and it reuses scratch counting buffers instead of rebuilding arrays every pan/zoom. On the data-loading side, [get_scatter_data.js](/Users/feni/Documents/celldega/js/read_parquet/get_scatter_data.js) now flattens coordinates in one allocation instead of repeated array growth, and [row_group_tile_reader.js](/Users/feni/Documents/celldega/js/read_parquet/row_group_tile_reader.js) adds smarter row-group reads and request caching.

Why that speeds things up:
- Less allocation: fewer `{name, x, y}` objects means less GC pressure.
- Less string work: counting by integer gene/category IDs is much cheaper than repeatedly comparing names.
- Less repeated scanning: the old code filtered transcripts/cells, then reduced them, then did repeated `find` calls; the new code counts directly into reusable arrays.
- Less redundant rendering work: if the tiles in view have not changed, it skips refetching/rebuilding transcript and path data.
- Less UI churn: bar-chart data is only published when it actually changed.

There’s also a preprocessing/scalability improvement in [trx_tile.py](/Users/feni/Documents/celldega/src/celldega/pre/trx_tile.py): large transcript datasets can now be tiled with a disk-backed streaming path instead of holding everything in memory. That’s less about making a single frame render faster and more about making big Atera/Xenium-style datasets practical to preprocess and load without memory blowups.

So at a high level: the speedup comes from doing the same work with smaller data structures, fewer passes, and more caching. The user-visible effect should be smoother close-up panning/zooming, faster transcript updates, and less lag when bar summaries refresh.

cornhundred:
Additional Context (ChatGPT)

The most complex files are the ones that changed both data shape and execution flow:

* `js/deck-gl/core/calc_viewport.js`: this is where close-up behavior, viewport bar aggregation, tile-change caching, and update throttling all come together.
* `js/vector_tile/transcripts/grab_trx_tiles_in_view.js`: this is the core transcript-loading refactor, changing transcripts from name/object-heavy data into compact buffers with gene IDs.
* `src/celldega/pre/trx_tile.py`: this is the biggest preprocessing change, adding the disk-backed streaming path for large transcript tiling jobs.
* `js/read_parquet/row_group_tile_reader.js`: this adds smarter row-group reading, caching, and multi-file handling, which supports the faster loading path.
* `js/viz/yearbook.js`: same compact-buffer/counting strategy got propagated into Yearbook, so it is fairly dense too.

The 4 highest-level changes in this request were:

* Transcript and cell data were reworked into compact typed-array structures instead of arrays of JS objects. That change is spread across `grab_trx_tiles_in_view.js`, `compact_data.js`, `trx_layer.js`, and `cell_layer.js`.
* Close-up viewport updates were optimized so gene and cell bar summaries are counted with reusable buffers and only recomputed or republished when needed. That mostly lives in `calc_viewport.js` and `yearbook.js`.
* Parquet row-group reading became more resilient and efficient, with request caching plus metadata/layout fallbacks for chunked datasets. That work is in `row_group_tile_reader.js`, `cbg_row_group_reader.js`, and `image_row_group_reader.js`.
* Preprocessing was extended for larger and newer Xenium/Atera-style datasets, especially by adding streaming transcript tiling and more flexible morphology file discovery. That is centered in `trx_tile.py`, `image_info.py`, and `src/celldega/pre/__init__.py`.

## “array of objects” layout to a “structure of arrays” layout.

Before, transcript data looked roughly like this:

```js
[
  { name: 'EPCAM', x: 101.2, y: 55.8 },
  { name: 'KRT19', x: 102.7, y: 56.1 },
  { name: 'EPCAM', x: 103.0, y: 57.4 }
]
```

Now it looks more like this:

```js
{
  geneIds: Int32Array([12, 44, 12]),
  positions: Float64Array([101.2, 55.8, 102.7, 56.1, 103.0, 57.4]),
  size: 2
}
```

For cells, it uses the same idea: numeric position buffers plus integer category IDs, with a separate small lookup list for category names.

Why this is better:
- Arrays of JS objects are expensive. Each transcript becomes its own object with property lookups like `.name`, `.x`, and `.y`.
- Typed arrays are compact contiguous numeric buffers, so the browser can iterate them faster and with much less garbage collection.
- Gene names are no longer repeated on every transcript. Each transcript stores a small integer ID instead, and the real gene name is only looked up when needed.
- Counting gets much cheaper. Instead of filtering objects and comparing strings, the code can just do “read `geneId`, increment `geneCounts[geneId]`”.

So the high-level translation is:

- Old way: “for each transcript, store a mini record object”
- New way: “store one big numeric buffer for positions and one big numeric buffer for IDs”

That reduces memory overhead, avoids repeated string comparisons, and turns the close-up aggregation work into fast numeric array processing.

The front-end changes were not just “store data differently”; they also updated the code that consumes it so it can actually render and interact with that compact format:

Transcript loading now builds compact numeric buffers instead of per-transcript objects.
The transcript layer reads positions from those buffers and uses integer gene IDs for coloring/filtering.
Tooltip/name lookup was updated to turn those gene IDs back into display names when needed.
Close-up gene bars and Yearbook summaries were updated to count directly from the compact buffers.
One nuance: this is most true for transcripts. For cells, the compact format is mainly used for fast aggregation and bar-chart updates, while the main visible cell layer still largely uses the existing cell scatter objects for rendering.

So high-level: yes, the frontend was updated to work with the compact representation where it matters for the speedup, especially in transcript rendering and close-up summary calculations.

PR #300, opened 2026-04-27
DEGA-266-yb-sample

cornhundred:


cornhundred:
# PR #300 Roadmap: AnnData selection & sampling layer for Yearbook

Source: https://github.com/broadinstitute/celldega/pull/300

Adds a Python-side query/sampling layer over `AnnData` and wires it into `Yearbook`, so a stable, serializable list of cell IDs — produced by a filter and a sampler — drives portrait ordering. A parallel stateless browser query covers the no-`AnnData` case.

## The core idea
`Yearbook` renders a grid of cell portraits and needs **an ordered list of cell IDs** to do it. Today that list is assembled ad hoc. This PR separates *which cells, in what order, and why* from the widget: a small composable API turns an `AnnData` into a `Selection` (ordered IDs + provenance), and the widget just paginates over it.

The unit model: **`Attribute`** (a lazy value reference) → **`Query`** (defines the candidate set) → **`Sampler`** (orders/subsets the candidates) → **`Selection`** (ordered IDs + the query/sampler/scores/provenance that produced them). Each stage is independently serializable, so a selection made in Python round-trips to the frontend as JSON.

## 1. `dega.select` module — the query/sampling layer
`src/celldega/select/core.py` (~1,200 lines, new)
- **`Selector`** — bound to one `AnnData`. Entry points `attr(name)` (an `adata.obs` column) and `gene(name, layer=, raw=)` (an expression vector from `X`, a named layer, or `.raw`). `select(query=, sampler=)` returns a `Selection`.
- **`Attribute`** — lazy reference resolved only at evaluation time; sparse-safe (`toarray`/`todense` → 1-D).
- **Query algebra** — `PredicateQuery` (`==`, `!=`, `<`/`<=`/`>`/`>=`, `.isin()`, `.notin()`, `.isna()`, `.notna()`, `.between()`) composed with `&` / `|` / `~` into `BooleanQuery`. NaNs resolve to `False`.
- **Five samplers** — `RandomSampler`, `RankSampler` (top/bottom by value), `QuantileBinSampler` (low/mid/high region, with `proportion`/`percentile` shortcuts), `GaussianSampler` (weight toward a target value), `StratifiedSampler` (even per-category or a total quota). All seedable; ranking samplers attach `scores`. Constructed via the `selector.samplers.*` namespace.
- **`Selection`** — list-like (iterate/index/slice), plus `names()`, `to_dataframe()`, `to_json()`, `page()`, and `candidate_count`/`selected_count`/`provenance`.
- **Safe preview guard** — if a query matches more than `default_preview_n` (1,000) entities and no sampler is given, `select()` returns a *deterministic random preview* and warns, instead of dumping the full set. `sampler="all"` opts into everything; `sampler=3000` is integer shorthand for a seeded random sample.

## 2. Public API export
`src/celldega/select/__init__.py` (new) + `src/celldega/__init__.py`
- Exposes the `select` classes and adds `select` to the top-level namespace, so usage is `dega.select.Selector(adata)`. Adds a `__dir__()` helper for cleaner interactive discovery.

## 3. Yearbook integration
`src/celldega/viz/widget.py`
- `Yearbook(selection=...)` accepts a `Selection`, a JSON-like selection dict, or a plain ID list, normalized by the new **`_selection_to_payload()`** helper.
- When `selection=` is given, Yearbook uses its IDs as the ordered `cells` list, stores the full payload for provenance, resets `current_page` to 0, and rejects passing both `selection` and `cells`.
- `cells=` remains a first-class plain-list input for the "I already know the IDs" case.

## 4. Front-end vs. back-end selection (the `query` → `front_end_query` rename)
`src/celldega/viz/widget.py`, `js/celldega.js`, `js/viz/yearbook.js`
- The Yearbook trait formerly named **`query`** is renamed to **`front_end_query`** to disambiguate the *stateless, in-browser* query (evaluated against LandscapeFiles, **no Python/`AnnData` needed**) from the Python `select` query module. `query=` still works but emits a `DeprecationWarning` mapping to the new name.
- `front_end_query` is a deliberately narrow counterpart: a single cluster filter and/or single-gene ranking (`{"gene": "BRCA1"}`, `{"cluster": {...}}`, `{"max_cells": N}`). A single-gene front-end query is the in-browser equivalent of the back-end `rank` sampler.
- The JS bundle (`docs/assets/js/celldega.js`) is regenerated to pick up the trait rename.

## 5. Correctness hardening (from review)
`src/celldega/select/core.py`
- **Quantile bins now partition** — interior boundaries are half-open (`low ≤ low_cut`, `mid (low_cut, high_cut)`, `high ≥ high_cut`) so tie-heavy count data no longer double-assigns a value to two bins.
- **Duplicate-name guards** — `Selector.__init__` raises on non-unique `obs_names` (they're the selection IDs); `gene()` raises lazily on a duplicated `var_names` entry only when that gene is actually queried, so duplicate gene symbols don't block obs-only work.

## 6. Docs, notebook & tests
- New `docs/python/select/api.md` (concepts, samplers, preview guard, validation, Yearbook integration) with a **"Backend vs. Front-End Selection"** comparison table; `docs/python/viz/api.md` gains **"Yearbook Selection" + "Front-End Query"** sections documenting the three population paths and the `front_end_query` dict shapes. `docs/python/index.md`, `docs/python/api.md`, and `mkdocs.yaml` updated for the `select` nav entry.
- `notebooks/yb_sampling.ipynb` demonstrates the query → sample → Yearbook arc.
- Tests: the new select suite (**18 tests**) and Yearbook widget tests (**14**) pass — covering filters, gene queries, every sampler, preview warnings, `sampler="all"`, integer shorthand, pagination, serialization, missing-attribute/gene errors, bin disjointness, the duplicate-name guards, and the `front_end_query`/`query` rename.
- `pyproject.toml` adds `setuptools` to dev deps. Test-isolation stubs added in `test_nbhd/test_nbp.py` and `test_pre/test_chromium.py`; `test_pre/test_row_groups.py` is formatting-only.

## Suggested review order
1. **`select/core.py`** — the API surface and where correctness lives (query semantics, sampler edge cases, layer/raw handling, serialization, preview behavior).
2. **`tests/unit/test_select/test_selector.py`** — confirms the intended AnnData cases; weigh whether sparse/layer/raw/categorical coverage is sufficient.
3. **`viz/widget.py`** — the `selection=` path (page reset, provenance) and the `front_end_query` rename + deprecation.
4. **JS** (`celldega.js`, `viz/yearbook.js`) — only the trait-name reads/writes changed.
5. Docs + notebook — confirm the public examples match the API to adopt.
6. Test stubs / formatting-only changes — low priority; check the stubs don't mask real import problems.

## Known limitations to weigh before merge
- **Selection payload duplicates IDs on the wire** — the frontend renders from `cells`; the synced `selection` dict also carries `ids` (and scores), so they travel twice. Pure provenance today; trimmable if large-preview payload size matters.
- **`front_end_query` rename is a soft break** — the constructor alias covers `query=`, but not direct post-construction trait assignment (`yb.query = {...}`). An observed alias trait could close that gap if needed.
- **Stratified output is grouped, not interleaved** — IDs come out ordered by category, which may surprise a grid that expects interleaving.
- **`.docs-select-preview/` + `mkdocs.select-preview.yml`** — a tracked standalone preview of the select doc, now a stale duplicate of `docs/python/select/api.md`. Looks like a dev scaffold; recommend removing from the PR.
- **Terminology drift vs #303** — this branch's docs/comments say "LandscapeFiles"; #303 renames to "DegaFiles" (`path_dega_files`). Align with whichever merges first.
- **Coordination with #303** — no architectural clash (different widgets/modules), but expect mechanical merge conflicts in `__init__.py`, `viz/widget.py`, the doc nav files, and the generated `celldega.js` (resolve the bundle by rebuilding, not hand-merging).

cornhundred:
@jaspreetishar, we can focus our review on the select module and not the viz or JavaScript changes since we will iterate on the front-end visualization more later.

jaspreetishar (approved):
looks good to me. the notebook and the tests are running successfully.

PR #303, opened 2026-05-05
DEGA-260-Neighborhood-level-data-schema-and-class

jaspreetishar:


cornhundred:
# PR Roadmap: MuData-backed Collection API for higher-order biological entities

Establishes a new way Celldega models entities *above* the single-cell level, and wires them into the Landscape/Matrix visualizations.

## The core idea
Single-cell data comes off the instrument already constructed with feature spaces. **Datasets and neighborhoods don't** — Celldega has to both *construct the entity* (its observation axis) and *calculate its feature spaces*. This PR introduces a base class that captures that pattern and two concrete entity types built on it.

The unit model: **`AnnData` = one feature modality (matrix); `MuData` = one collection** (many modalities sharing an observation axis, plus relations and metadata).

## 1. `CelldegaCollection` — the base abstraction
`src/celldega/collection/collection.py` (~307 lines, new)
- Thin typed wrapper over a `MuData`. Its core accessors are **aliases** to the underlying MuData: `obs`→`mdata.obs`, `mod`→`mdata.mod`, `relations`→`mdata.obsp`, `uns`→`mdata.uns["celldega"]` (schema version, `collection_type`, `obs_entity_type`, provenance).
- **`_align_mod_to_obs`** — the single alignment path all modality-attachment funnels through (`add_mod`). Sparse-safe; zero-fills missing rows for count-like columns.
- `add_relation_modality` (materialize a relation as a clusterable modality), `write`/`read` (persist the MuData).

## 2. `DatasetCollection` — dataset/sample/patient entities
`src/celldega/dataset/collection.py` (~479 lines, new)
- "Construction" = **binning single-cell `adata` over `dataset_col`** to build the dataset obs axis (no geometry needed).
- Feature methods: `calc_dataset_by_pop` (dataset × population) and `calc_dataset_signature` (dataset × gene for one category value, with explicit `missing_datasets="nan"`).

## 3. `NeighborhoodCollection` + nbhd module restructure
`src/celldega/nbhd/collection.py` (~383 lines, new) and a heavily reworked `neighborhoods.py`
- Neighborhoods are **spatially constructed** (from a GeoDataFrame via `from_gdf`, or `alpha_shape`/`hextile` module functions), then carry feature modalities + relations + geometry.
- Public calc lives **on the class**, each a thin wrapper over a private spatial kernel in `neighborhoods.py`:
  - feature modalities: `calc_nbhd_by_pop`, `calc_nbhd_by_gene`
  - relations (→ `obsp`): `calc_nbhd_overlap`, `calc_nbhd_bordering`
  - obs metadata: `calc_nbhd_transcript_assignment` (adds `total_transcripts`, `unassigned_transcripts`, `transcript_assignment_proportion`)
- `drop_missing` flag controls whether sub-`min_cells` neighborhoods are dropped from the axis (default) or kept with zero-filled rows.
- **Geometry is a single source of truth** (`gdf`, with `geometry` as a read-only alias).

## 4. Visualization path (micron→pixel + Landscape/Matrix wiring)
- `NeighborhoodCollection` can hold a **micron-to-pixel transformation matrix** (`transformation_matrix=` ctor arg, `set_transformation_matrix`, or `load_transformation_matrix` from DegaFiles; round-trips via `uns`). `to_pixel_gdf()` emits a `geometry_pixel` column for rendering.
- `Landscape(nbhd=collection)` now accepts a `NeighborhoodCollection` directly (duck-typed `.geometry`/`.gdf`/`.obs`), and `Matrix` **auto-infers row/col entities** from the `var["entity_type"]` / `uns` our modalities write. *(Front-end fix merged via #304: `viz/widget.py`, `clust/matrix.py`, deck.gl `nbhd_layer.js`.)*

## 5. What was removed (intentional API breaks)
- **The legacy `NBHD` class** — retired entirely; its feature/relation logic moved onto `NeighborhoodCollection`, its construction is now module functions. No more "Neighborhood class distinct from NeighborhoodCollection."
- **`HierarchyResult`** — speculative clustering-metadata schema with no producer/consumer; deferred until the `Matrix → collection → viz` workflow exists.
- **`neighborhood_collections`** link on `DatasetCollection`, **`calc_nbhd_by_image`** (NBI), and the self-returning accessor shims — dropped as speculative/unused.
- Three drifted alignment helpers collapsed into the one base helper; `_relation_from_square_adata` reworked to be sparse-safe (no dense round-trip).

## 6. Supporting renames, docs & tests
- `path_landscape_files → path_dega_files` / `LandscapeFiles → DegaFiles` across `pre/`.
- New docs: `collection/api.md` (incl. the explicit `relations` = `mdata.obsp` alias note), `dataset/api.md`; updated `python/index.md`, `mkdocs.yaml`.
- Example notebooks show the full arc: hextile construction → `calc_nbhd_by_pop` + `calc_nbhd_by_gene` (signature) → Matrix/Clustergram → `Landscape(nbhd=collection)`; the DatasetCollection notebook covers population + signature.
- **64 unit tests pass** (incl. the merged viz tests; an old cross-directory test-isolation failure and a stale `celldega.collections` import in the viz test were fixed along the way).

## 7. Developer environment (uv-based setup)
`scripts/setup.sh`, `scripts/config.sh`, `scripts/utils.sh`
- The dev env (`dega/`) is now built with **uv on a standalone CPython** (`uv venv --python-preference only-managed`) — never Anaconda's interpreter.
- **Why:** a venv created from Anaconda's Python drags in Anaconda's native libs (e.g. `libgio`), which collide with the GDAL/GLib bundled by the geo wheels (`geopandas`/`pyogrio`) and crash scanpy/geopandas at runtime (`GLib-GObject-CRITICAL` / hard kernel crash). A clean standalone base removes the clash.
- `setup.sh` now: requires `uv`, creates the env with uv (and **auto-recreates** an existing Anaconda-based `dega/`), installs via `uv pip install -e ".[dev]"`, and **registers a "Python (dega)" Jupyter kernel** so notebooks run on this env rather than Anaconda's base kernel. `--status` reports `uv`; `scripts/*.sh` made executable.
- **Follow-up (not in this PR):** add `requires-python` to `pyproject.toml` + a committed `uv.lock` (`uv sync`) for fully pinned, reproducible installs — the current `uv pip install -e ".[dev]"` is a drop-in for the old pip flow and is not yet lock-pinned.

## Suggested review order
1. **`collection/collection.py`** — the schema contract (alignment, relations, read/write). Everything else builds on it.
2. **`dataset/collection.py`** — cleanest; confirms the entity pattern.
3. **`nbhd/collection.py`** — the public neighborhood API (feature, relation, transcript-assignment, transformation-matrix methods).
4. **`nbhd/neighborhoods.py`** — private spatial kernels + the destructive `_subset_neighborhood_collection_to_obs` (highest-risk helper).
5. **`viz/widget.py` + `clust/matrix.py`** (from #304) — `_coerce_nbhd_for_landscape` and entity inference.
6. Tests, then docs/notebooks.
7. **`scripts/` (uv setup)** — independent dev-tooling change; quick read.

## Known limitations to weigh before merge
- **Round-trip gap:** geometry/`gdf` and `memberships` are in-memory only — lost on `write`/`read` (the `transformation_matrix` *does* now round-trip via `uns`).
- **Missing-vs-zero:** alignment zero-fills missing rows (except NaN-bearing signatures); fine for counts, worth noting for richer modalities.
- **Platform assumption:** `calc_nbhd_transcript_assignment` reads the precomputed `cell_id` from instrument transcripts and treats `"UNASSIGNED"` as the sentinel — that's **Xenium-specific** (errors if `cell_id` is missing, warns if the sentinel is absent).
- **Env reproducibility:** the uv setup fixes the Anaconda/GLib crash but isn't lock-pinned yet (see §7 follow-up).
- Two near-duplicate clustering cells remain in the neighborhood notebook (from #304) — candidate to trim to one.

cornhundred (approved):
I reviewed the code and ran the test notebook. I think it's ok to merge in, continue testing on real data, and iterate if needed.

docs/python/index.md
```diff
@@ -51,30 +32,96 @@ mat.cluster()
...
+dataset-level and neighborhood-level data:
+
+- `dega.dataset.DatasetCollection` for dataset, sample, tissue section, or patient observations
+- `NeighborhoodCollection` for neighborhood or spatial-region observations
+
+```python
+import celldega as dega
+
+dset = dega.dataset.DatasetCollection(adata, dataset_col="sample_id")
+nbhd = dega.NeighborhoodCollection(obs=neighborhood_obs, geometry=neighborhood_gdf)
```
jaspreetishar: correction: `dega.nbhd.NeighborhoodCollection()`

src/celldega/dataset/collection.py
```diff
@@ -0,0 +1,548 @@
...
+    """
+    resolved_dataset_col = _resolve_dataset_col(dataset, dataset_col or dataset.dataset_col)
+    modality = _population_modality_from_adata(
+        adata,
+        dataset_col=resolved_dataset_col,
+        category=category,
+        output=output,
+        min_cells=min_cells,
+    )
+    dataset.add_mod(modality_name, modality, var_entity_type="cell_population")
```
jaspreetishar, line 548: can we avoid hardcoding "cell_population" for var_entity_type? we can add var_entity_type as an additional parameter in _calc_dataset_by_pop()

jaspreetishar (commented):
looks great! I have a couple of minor comments. The notebooks and pytests are running successfully.

src/celldega/nbhd/collection.py
```diff
@@ -0,0 +1,529 @@
...
+            adata,
+            self.gdf,
+            category=category,
+            nbhd_col=self.nbhd_col,
+            min_cells=min_cells,
+            output=output,
+        )
+        if drop_missing:
+            _subset_neighborhood_collection_to_obs(self, pd.Index(modality.obs_names.astype(str)))
+        self.add_mod(modality_name, modality, var_entity_type="cell_population")
```
jaspreetishar, line 335: https://github.com/broadinstitute/celldega/pull/303#discussion_r3430951929: same comment here

docs/python/index.md
cornhundred: Thanks, NeighborhoodCollection is also exposed at the base level (we can change this later), but lets guide users to access it from the submodule.

src/celldega/dataset/collection.py
cornhundred: We actually do want to hard code this since the var-axis entity is always `cell_population` from this method. The user can give a custom name for their modality using `modality_name`. Our other feature-calculation methods also hardcode their entity type (e.g., `calc_nbhd_by_gene` hardcodes `gene`).

src/celldega/nbhd/collection.py
cornhundred: Same, we want to hardcode the var_entity_type.

PR #308, opened 2026-06-28
Draw by drag

cornhundred:


cornhundred:
## Draw neighborhoods by dragging (freehand SKTCH mode)

Replaces the click-to-place-vertices polygon tool with freehand drag-to-draw,
making neighborhood sketching faster and more natural, and adds keyboard
shortcuts to enter/exit sketch mode.

### Changes
- **Freehand drawing** — SKTCH now uses `DrawPolygonByDraggingMode` instead of
  `DrawPolygonMode`; drag to trace a region, release to finish. Sampling is
  throttled to ~60fps (`throttleMs: 16`).
- **Keyboard shortcuts** — press `s` to enter sketch mode and `Escape` to exit.
  The shortcut is scoped to whichever widget the pointer is hovering (checked
  via `:hover` at keydown time), so multiple Landscape widgets on one page
  (e.g. a notebook) don't all toggle at once. Guarded against modifier keys
  and input/editable targets so it won't fire while typing, and the listener
  is registered in the capture phase and calls `stopPropagation()` when it
  acts, so it claims the keystroke before it can reach host-page shortcuts
  (e.g. Jupyter's own `s`-to-save binding in command mode). Sketch mode is
  single-shot — a completed stroke auto-returns to view mode.
- **Refactor** — extracted the SKTCH enter/exit logic into a single
  `set_sketch_mode()` helper shared by the button and the keyboard shortcut,
  and made the NBHD-edit-visibility toggle (which forces the layer/button
  back to view state independently of that helper) reset `viz_state.edit.mode`
  in lockstep, so the two can't drift out of sync.
- **Lifecycle** — the keydown listener is torn down in `landscape_ist.finalize()`,
  and `make_ist_ui_container` always cleans up any prior listener on the same
  `viz_state` before conditionally re-registering (not just when NBHD edit is
  currently on), so re-running it can't orphan a handler.
- **Perf** — `addTentativePosition` / `cancelFeature` edits short-circuit to a
  layer refresh, skipping add-feature finalization during a drag; that refresh
  now passes `viz_state` through to `get_layers_list` like every other call
  site, so point-cloud layer filtering stays correct on that path.

### Files
- `js/ui/ui_containers.js` — `set_sketch_mode`, shortcut handler, hover-scoping,
  listener cleanup
- `js/deck-gl/layers/edit_layer.js` — `modeConfig` support, tentative-position fast path
- `js/viz/landscape_ist.js` — `edit.mode` kept in sync on NBHD-edit toggle, shortcut cleanup on finalize

### Notes
- Merged `main` (0.18.0, set/collection work) into this branch; version bumped to **0.19.0**.
- Bundle `docs/assets/js/celldega.js` regenerated via `npm run build`.

### Known limitation / follow-up
- Mode values (`'sktch'`/`'view'`/`'modify'`) are bare string literals shared
  across files; consolidating them into shared constants is a pre-existing
  cleanup left for a separate PR.

PR #314, opened 2026-07-13
Align module

cornhundred:
Working on building an align module that will be used for various alignment tasks.


cornhundred:
## 2026-07-15 Summary

Adds `celldega.align`, a module for aligning spatial single-cell data in **physical** (spatial) coordinates. This PR implements serial 3D slice reconstruction (reference/atlas registration and cross-modality registration are separate, future work).

The pipeline is three composable, independently reusable steps:

1. **`calc_landmarks(adatas, cluster_key, slice_attr=None) -> DataFrame`** — one row per cluster label per slice (`label`, `x`, `y`, `count`). Accepts a single `AnnData`, a list of `AnnData`, or one combined `AnnData` + `slice_attr`.
2. **`calc_alignment_transform(landmarks, ...) -> SerialAlignmentTransform`** — fits a *spatial* transform per slice by chain-walking outward from a `reference` slice (never a single distant reference), and returns a reusable transform object. Touches only `landmarks`, no cell data, no Z.
3. **`align_serial_slices(adatas, transform, z_space=1.0, z_coord=None, key_added="Z") -> AnnData`** — applies a fitted transform to a specific set of `AnnData`, aligning `obsm["spatial"]` and assigning Z.

Z assignment lives in step 3, not step 2 — it has no effect on the spatial fit, so the identical fitted transform can be applied with different Z choices without refitting anything.

## API

```python
import celldega as dega

adatas = [adata_slice0, adata_slice1, adata_slice2]
landmarks = dega.align.calc_landmarks(adatas, "leiden")

transform = dega.align.calc_alignment_transform(
    landmarks,
    alignment_window=2,               # pool landmarks from up to 2 prior neighbors
    method="tps",                     # or "procrustes" (default) for a rigid fit
    weight_by_adjacent_counts=True,   # default: trust larger clusters more
)

adata_aligned = dega.align.align_serial_slices(adatas, transform, z_space=10.0)  # microns between sections
adata_aligned.obs["Z"]
adata_aligned.uns["align_serial_slices"]   # fit params + landmark provenance

# The fitted transform is a first-class, reusable object:
transform.apply_to_points(slice_id, other_points)   # e.g. segmentation polygons, transcripts
transform.save("path/to/dir")                        # plain files, no pickle
reloaded = dega.align.SerialAlignmentTransform.load("path/to/dir")
```

- **`method="procrustes"`** (default): rigid rotation + translation, scaling always off (physical sections shouldn't need rescaling). **`method="tps"`**: non-rigid thin-plate-spline warp, for deformation a rigid fit can't capture.
- **`weight_by_adjacent_counts`**: weights each shared landmark by the geometric mean of its cell count in the current slice and neighbor window; neutral fallback for landmarks with no count (e.g. manually placed).
- **`alignment_window`**: registers each slice against a window of already-aligned neighbors rather than just the immediately-previous one, reducing sensitivity to any one neighbor's noise.
- **`SerialAlignmentTransform`**: holds one `Transform` per slice (reference is identity) and fit provenance (`landmarks_initial`, `landmarks_aligned`, per-slice diagnostics including leave-one-out residuals) — no Z, see above. Also:
  - `.apply_to_points(slice_id, points)` — reuse the fitted alignment on any `(n, 2)` point array tied to that slice.
  - `.save(path)` / `SerialAlignmentTransform.load(path)` — a plain directory (`metadata.json`, `transform_log.json`, `landmarks_initial.parquet`, `landmarks_aligned.parquet`, `transforms/<slice_id>.npz`); no `pickle` involved, including for the TPS spline (persisted as its plain-array fit inputs and rebuilt on load, not the fitted object itself).

## Scope

**Deferred:** reference/atlas registration, cross-modality (image-based) registration, the interactive point-drawing widget for manual landmarks (`align_serial_slices` already consumes a manually-built landmarks table in the same shape `calc_landmarks` produces — only the drawing tool itself is deferred), and storing landmarks in `NeighborhoodCollection` (plain `DataFrame` for now).

## Files

- `src/celldega/align/_transform.py` (340 lines) — `Transform` protocol, `fit_transform_procrustes`, `fit_transform_tps`, `leave_one_out_residuals`, `save_transform`/`load_transform`.
- `src/celldega/align/landmarks.py` (97 lines) — `calc_landmarks`.
- `src/celldega/align/_slices.py` (80 lines) — shared multi-slice input resolution.
- `src/celldega/align/serial_slices.py` (588 lines) — `SerialAlignmentTransform`, `calc_alignment_transform`, `align_serial_slices`.
- `src/celldega/align/__init__.py` (75 lines) — public exports + roadmap docstring.
- `tests/unit/test_align/` — `test_transform.py` (223 lines), `test_serial_slices.py` (589 lines), `test_landmarks.py` (101 lines); 54 tests total.
- `docs/python/align/api.md`, `mkdocs.yaml` — docs nav entry.
- `src/celldega/__init__.py` — exports `dega.align`.

## Tests

54 tests covering: exact-transform recovery and reference-is-identity; `z_space`/`z_coord` (including reusing one fitted transform with two different Z schemes); `method="tps"` vs. `"procrustes"` on a non-affine warp; the never-rescales guarantee; weighted-fit correctness; `alignment_window` reducing compounded error; leave-one-out residuals and their opt-out; `calc_landmarks` across all three input modes; semi-manual landmark appending; `align_serial_slices`'s slice-order-mismatch validation; `apply_to_points` reuse on arbitrary points; and `.save()`/`.load()` round-trips for both Procrustes and TPS (confirming the TPS spline specifically survives and still interpolates exactly).

src/celldega/align/landmarks.py
```diff
@@ -0,0 +1,97 @@
...
+def _one_slice_landmarks(adata: AnnData, cluster_key: str) -> pd.DataFrame:
+    if cluster_key not in adata.obs.columns:
+        raise ValueError(f"'{cluster_key}' is not a column in adata.obs")
+    spatial = adata.obsm.get("spatial")
+    if spatial is None or np.asarray(spatial).shape[1] < 2:
+        raise ValueError("adata must have obsm['spatial'] with at least 2 columns (x, y)")
+
+    xy = np.asarray(spatial)[:, :2]
+    df = pd.DataFrame(xy, columns=["x", "y"], index=adata.obs_names)
+    df["label"] = adata.obs[cluster_key].astype(str).to_numpy()
```
jaspreetishar: Could we exclude cells with NaN cluster labels before calculating the landmarks, as they might silently pollute the alignment fit with "unclustered cells" pseudo-cluster.

src/celldega/align/serial_slices.py
```diff
@@ -0,0 +1,588 @@
...
+    aligned_slices = []
+    for slice_id, adata in zip(slice_ids, slices, strict=True):
+        spatial = np.asarray(adata.obsm["spatial"], dtype=float).copy()
+        spatial[:, :2] = transform.apply_to_points(slice_id, spatial[:, :2])
+        adata.obsm["spatial"] = spatial
+        adata.obs[key_added] = z_values[slice_id]
+        adata.obs[slice_attr] = slice_id
+        aligned_slices.append(adata)
+
+    adata_aligned = ad.concat(aligned_slices, join="outer")
```
jaspreetishar, line 587: Could we preserve unique obs_names during concatenation, for example by using a slice-specific suffix or validating that the indices are globally unique?

src/celldega/viz/widget.py
```diff
@@ -483,6 +528,41 @@ def _reset_index_for_parquet(df):
...
+                # MB) — when base_url is a local dev server, write a small
+                # sidecar file next to it instead and let the frontend fetch it
+                # over HTTP, exactly like the base cell_metadata.parquet. Falls
+                # back to the comm-synced bytes trait otherwise (fine for
+                # smaller datasets or non-local base_urls).
+                base_url_str = kwargs.get("base_url") or ""
+                local_dir = _local_dir_for_url(base_url_str)
+                if local_dir is not None:
+                    cache_name = f".celldega_centroids_{uuid.uuid4().hex[:10]}.parquet"
+                    centroid_df.to_parquet(local_dir / cache_name, index=False)
```
jaspreetishar: Could we add a cleanup strategy for these temporary centroid files so stale .celldega_centroids_*.parquet files do not accumulate?

src/celldega/viz/widget.py
```diff
@@ -483,6 +528,41 @@ def _reset_index_for_parquet(df):
                 )
                 pq_umap = _df_to_bytes(umap_df)
 
+            if use_adata_3d_centroids and "spatial" in adata.obsm:
```
jaspreetishar: Could we make use of `technology == "point-cloud"` here instead of introducing the `use_adata_3d_centroids` flag? A plain 2D Landscape(adata=..., technology="Xenium") still pays the serialization/disk-write cost for data nobody renders. Driving it off technology directly would make the Python guard match the documented behavior instead of relying on the frontend to silently discard the work.

src/celldega/align/landmarks.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+__all__ = ["calc_landmarks"]
+
+
+def _one_slice_landmarks(adata: AnnData, cluster_key: str) -> pd.DataFrame:
+    if cluster_key not in adata.obs.columns:
+        raise ValueError(f"'{cluster_key}' is not a column in adata.obs")
+    spatial = adata.obsm.get("spatial")
+    if spatial is None or np.asarray(spatial).shape[1] < 2:
```
jaspreetishar, line 36: Align's coordinate extraction only checks obsm["spatial"], not the obs x/y-centroid fallbacks that _cell_coords (src/celldega/set/collection.py) and _points_from_reference (src/celldega/nbhd/gradient.py) already support. Maybe it's worth reusing one of those instead of a third copy.

jaspreetishar (changes requested):
Looks great overall. My main suggestions are to handle NaN landmarks, preserve unique obs_names, restrict centroid overrides to point-cloud views, clean up temporary centroid files, and reuse the existing coordinate-extraction helpers so the align module also supports the same obs x/y-centroid fallbacks.

EDIT: added one more comment related to `centroids_parquet/centroids_url`

src/celldega/viz/widget.py
```diff
@@ -282,17 +317,25 @@ class Landscape(anywidget.AnyWidget):
...
+    use_adata_3d_centroids = traitlets.Bool(True).tag(sync=True)
+    centroids_url = traitlets.Unicode("").tag(sync=True)
+
     def __init__(self, **kwargs):
         adata = kwargs.pop("adata", None) or kwargs.pop("AnnData", None)
         pq_meta_cell = kwargs.pop("meta_cell_parquet", None)
         pq_meta_cluster = kwargs.pop("meta_cluster_parquet", None)
         pq_umap = kwargs.pop("umap_parquet", None)
         pq_meta_nbhd = kwargs.pop("meta_nbhd_parquet", None)
+        pq_centroids = kwargs.pop("centroids_parquet", None)
```
jaspreetishar: Passing centroids_parquet/centroids_url explicitly doesn't work when adata is also given - it gets silently overwritten by centroids computed from adata at line 531. Possible fix: only run the auto-computed centroids block when the caller hasn't already supplied one, e.g. by adding `and not pq_centroids and not centroids_url` to the if condition at line 531.

src/celldega/align/landmarks.py
cornhundred: Addressed in the latest commit.

src/celldega/viz/widget.py
cornhundred: Point-cloud gating (viz/widget.py) — centroid computation/serialization now requires technology to actually be a point-cloud variant, not just the use_adata_3d_centroids flag.

src/celldega/align/landmarks.py
cornhundred: Coordinate source — confirmed obsm['spatial'] should stay align's only source of truth; no change needed. I think we can move to use obsm['spatial'] as our source of truth and remove redundancies later.

src/celldega/viz/widget.py
cornhundred: We'll clean this up so that we only pass in the data from AnnData.

src/celldega/viz/widget.py
cornhundred: These temporary files are a workaround. The comms can break when >5-10MB of data is sent at once (e.g., millions of cell centroid positions in xyz) and we were experimenting with temporary files that effectively took the place of DegaFiles. I am just reverting to requiring DegaFiles.

cornhundred:
## Summary

Adds `celldega.align`, a module for aligning spatial single-cell data in **physical** (spatial) coordinates. This PR implements serial 3D slice reconstruction (reference/atlas registration and cross-modality registration are separate, future work).

The pipeline is three composable, independently reusable steps:

1. **`calc_landmarks(adatas, cluster_key, slice_attr=None) -> DataFrame`** — one row per cluster label per slice (`label`, `x`, `y`, `count`). Cells with a `NaN` cluster label (unclustered/QC-filtered cells) are excluded rather than pooled into a spurious `"nan"` landmark. Accepts a single `AnnData`, a list of `AnnData`, or one combined `AnnData` + `slice_attr`.
2. **`calc_alignment_transform(landmarks, ...) -> SerialAlignmentTransform`** — fits a *spatial* transform per slice by chain-walking outward from a `reference` slice (never a single distant reference), and returns a reusable transform object. Touches only `landmarks`, no cell data, no Z.
3. **`align_serial_slices(adatas, transform, z_space=1.0, z_coord=None, key_added="Z", cell_name_prefix=False) -> AnnData`** — applies a fitted transform to a specific set of `AnnData`, aligning `obsm["spatial"]` and assigning Z. Warns if the concatenated result's `obs_names` collide (common when slices reuse the same per-slice barcode convention); pass `cell_name_prefix=True` to prefix each cell with its slice id first, matching `celldega.viz.widget.Landscape`'s existing `cell_name_prefix` convention.

Z assignment lives in step 3, not step 2 — it has no effect on the spatial fit, so the identical fitted transform can be applied with different Z choices without refitting anything.

## API

```python
import celldega as dega

adatas = [adata_slice0, adata_slice1, adata_slice2]
landmarks = dega.align.calc_landmarks(adatas, "leiden")

transform = dega.align.calc_alignment_transform(
    landmarks,
    alignment_window=2,               # pool landmarks from up to 2 prior neighbors
    method="tps",                     # or "procrustes" (default) for a rigid fit
    weight_by_adjacent_counts=True,   # default: trust larger clusters more
)

adata_aligned = dega.align.align_serial_slices(
    adatas, transform, z_space=10.0,  # microns between sections
    cell_name_prefix=True,            # keep obs_names unique across slices
)
adata_aligned.obs["Z"]
adata_aligned.uns["align_serial_slices"]   # fit params + landmark provenance

# The fitted transform is a first-class, reusable object:
transform.apply_to_points(slice_id, other_points)   # e.g. segmentation polygons, transcripts
transform.save("path/to/dir")                        # plain files, no pickle
reloaded = dega.align.SerialAlignmentTransform.load("path/to/dir")
```

- **`method="procrustes"`** (default): rigid rotation + translation, scaling always off (physical sections shouldn't need rescaling). **`method="tps"`**: non-rigid thin-plate-spline warp, for deformation a rigid fit can't capture.
- **`weight_by_adjacent_counts`**: weights each shared landmark by the geometric mean of its cell count in the current slice and neighbor window; neutral fallback for landmarks with no count (e.g. manually placed).
- **`alignment_window`**: registers each slice against a window of already-aligned neighbors rather than just the immediately-previous one, reducing sensitivity to any one neighbor's noise.
- **`cell_name_prefix`**: keeps `obs_names` unique post-concatenation without inventing a new convention — reuses the same prefix-then-underscore scheme `Landscape(cell_name_prefix=True)` already knows how to strip when mapping cells back to per-slice LandscapeFiles.
- **`SerialAlignmentTransform`**: holds one `Transform` per slice (reference is identity) and fit provenance (`landmarks_initial`, `landmarks_aligned`, per-slice diagnostics including leave-one-out residuals) — no Z, see above. Also:
  - `.apply_to_points(slice_id, points)` — reuse the fitted alignment on any `(n, 2)` point array tied to that slice.
  - `.save(path)` / `SerialAlignmentTransform.load(path)` — a plain directory (`metadata.json`, `transform_log.json`, `landmarks_initial.parquet`, `landmarks_aligned.parquet`, `transforms/<slice_id>.npz`); no `pickle` involved, including for the TPS spline (persisted as its plain-array fit inputs and rebuilt on load, not the fitted object itself).

## Scope

**Deferred:** reference/atlas registration, cross-modality (image-based) registration, the interactive point-drawing widget for manual landmarks (`align_serial_slices` already consumes a manually-built landmarks table in the same shape `calc_landmarks` produces — only the drawing tool itself is deferred), and storing landmarks in `NeighborhoodCollection` (plain `DataFrame` for now). We also explored letting `Landscape` preview an aligned `AnnData` without writing real LandscapeFiles (comm-synced/on-the-fly); decided against it, so previewing still goes through the normal `celldega.pre` route.

## Files

- `src/celldega/align/_transform.py` (340 lines) — `Transform` protocol, `fit_transform_procrustes`, `fit_transform_tps`, `leave_one_out_residuals`, `save_transform`/`load_transform`.
- `src/celldega/align/landmarks.py` (97 lines) — `calc_landmarks`.
- `src/celldega/align/_slices.py` (80 lines) — shared multi-slice input resolution.
- `src/celldega/align/serial_slices.py` (588 lines) — `SerialAlignmentTransform`, `calc_alignment_transform`, `align_serial_slices`.
- `src/celldega/align/__init__.py` (75 lines) — public exports + roadmap docstring.
- `tests/unit/test_align/` — `test_transform.py` (223 lines), `test_serial_slices.py` (589 lines), `test_landmarks.py` (101 lines); 57 tests total.
- `docs/python/align/api.md`, `mkdocs.yaml` — docs nav entry.
- `src/celldega/__init__.py` — exports `dega.align`.

## Tests

57 tests covering: exact-transform recovery and reference-is-identity; `z_space`/`z_coord` (including reusing one fitted transform with two different Z schemes); `method="tps"` vs. `"procrustes"` on a non-affine warp; the never-rescales guarantee; weighted-fit correctness; `alignment_window` reducing compounded error; leave-one-out residuals and their opt-out; `calc_landmarks` across all three input modes, including excluding NaN-labeled cells; semi-manual landmark appending; `align_serial_slices`'s slice-order-mismatch validation; `apply_to_points` reuse on arbitrary points; `.save()`/`.load()` round-trips for both Procrustes and TPS (confirming the TPS spline specifically survives and still interpolates exactly); and `align_serial_slices` warning on duplicate `obs_names` while `cell_name_prefix=True` keeps them unique and silences it.

js/celldega.js
```diff
@@ -43,16 +43,39 @@ const render_landscape_ist = async ({ model, el }) => {
...
+  const centroidsUrl = model.get('centroids_url') || '';
+  if (centroidsUrl) {
+    // Millions of per-cell centroids don't fit through the widget comm
+    // channel, so they're served as a plain file (same as cell_metadata.parquet)
+    // instead of synced through model state — see `centroids_url` in widget.py.
+    const response = await fetch(centroidsUrl);
+    const centroidsBytes = new Uint8Array(await response.arrayBuffer());
+    centroids_data = (await objects_from_parquet(centroidsBytes, 'cell_id'))
+      .result;
+  } else {
```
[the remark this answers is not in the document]
jaspreetishar: If we end up removing `centroids_url` and `pq_centroids`, then I think this suggestion can be skipped.

js/celldega.js
```diff
@@ -163,9 +188,13 @@ const render_yearbook = async ({ model, el }) => {
...
 
+  const clusterAttr = model.get('cluster_attr') || 'leiden';
   const metaClusterBytes = model.get('meta_cluster_parquet');
   if (metaClusterBytes && metaClusterBytes.byteLength > 0) {
-    meta_cluster_data = await objects_from_parquet(metaClusterBytes, 'leiden');
+    meta_cluster_data = await objects_from_parquet(
+      metaClusterBytes,
+      clusterAttr
+    );
   }
```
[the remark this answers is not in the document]
jaspreetishar: I second this: the fix is to make Yearbook mirror what Landscape already does correctly - add `cluster_attr = traitlets.Unicode("leiden").tag(sync=True)`, and set `self.cluster_attr = cluster_attr` in Yearbook.__init__; both in widget.py. Once the real value is actually synced to the frontend, the existing `leiden` fallback should work fine as is.

docs/gallery/gallery_mouse_cranium_e14.md
```diff
@@ -0,0 +1,21 @@
...
+publicly yet. Before loading this page, serve the DegaFiles directory built by
+`StrataMap.ipynb` (`data/michal_landscape_files/E14_62_together_raw_v2_point-cloud/`)
+with a CORS-enabled static file server, e.g.:
+
+```bash
+npx http-server notebooks/data/michal_landscape_files/E14_62_together_raw_v2_point-cloud -p 8080 --cors
+```
+
+Then update `base_url` in `docs/assets/js/gallery_mouse_cranium_e14.js` to point at
+that server (default assumes `http://localhost:8080`).
```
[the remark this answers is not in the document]
jaspreetishar: The local URL will be removed once we switch to the hosted/ online URL like the other gallery examples.

js/celldega.js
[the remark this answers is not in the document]
cornhundred: This will be addressed in an upcoming commit.

cornhundred (commented):
Made the requested changes and will merge into main.

PR #323, opened 2026-07-29
Clustergram variants

cornhundred:
This pull request adds 
* a dot plot functionality to the Clustergram widget
* a new Composition widget for visualizing entity (e.g., dataset) composition by attribute (e.g., cell type) using stacked bar charts


cornhundred:
## Summary

Adds a new way to encode a matrix cell's value in `Clustergram` — **`dotplot`**
— alongside the existing `heatmap`, plus a new dedicated **`Composition`**
widget (a `Clustergram` subclass) for comparing category proportions/counts
across groups as stacked bars. Also restyles the shared control panel used by
both widgets, and includes a round of dendrogram/hover-interaction polish that
touches both `Clustergram` and `Composition`.

**Visual encodings (`Clustergram`)**
1. **`viz_mode` trait** — `"heatmap"` (opacity ∝ value, default) or `"dotplot"`
   (opacity from the main matrix, size from a *second* matrix — the classic
   dot-plot). Switching `viz_mode` on a live widget animates size + color
   together. (An earlier `"size"` mode — square size ∝ value alone — was
   considered and then removed before release: `dotplot` already covers "size
   encodes a value" via a proper secondary matrix, so there was no need to
   support both.)
2. **`Matrix.set_dot_matrix(dot)`** — attaches the secondary size-channel
   matrix, aligned to the main matrix **by row/col name** at export time
   (doesn't need to share clustered order). Accepts a `DataFrame`, array,
   `AnnData`, or another `Matrix`. (Fixed along the way: the `AnnData` branch
   wasn't transposing, silently zeroing out the aligned result.)
3. **`Matrix(collection=..., color_by=..., size_by=...)`** — builds the main
   matrix and attaches the size channel directly from a
   `SetCollection`/`DatasetCollection`, no manual `color_df`/`size_df`
   wrangling needed. `size_by` isn't limited to "fraction expressing" — any
   per-cell magnitude works (e.g. a significance score). `dot_plot=` is
   accepted as an alias for `size_by=`.
4. **`SetCollection.calc_signature(aggregate="fraction")`** — the "percent
   cells expressing" pseudobulk mode (binarizes at `expr_threshold`, skips
   library-size normalization since it's already `[0, 1]`) — pairs with a
   `mean` signature to drive dot size vs. color. `modality_name` is now a
   required argument on every `calc_signature` call (previously defaulted to
   `"expression"`/`"fraction"`/the feature type), so it's always explicit
   which modality a call produces.
5. **`TILE: PROP | UNIT`** control-panel toggle — switches dot size between
   fraction-encoded (default) and a full, unit-scaled tile, independent of the
   color/opacity channel, at any time.

**`Composition` widget**
6. **`dega.viz.Composition`** — a `Clustergram` subclass with
   `viz_mode="composition"`. Builds the matrix from a `DatasetCollection`/
   `SetCollection` population modality (or a raw `DataFrame`/`AnnData`); each
   group (dataset/sample) renders as a stacked bar, each population a colored
   segment. `viz_mode="composition"` is enforced as only settable via
   `Composition` (`TraitError` on a plain `Clustergram`) — the body was
   always designed to be reached through the dedicated widget.
7. **Stacked-bar rendering** — bars are bottom-anchored (standard bar-chart
   convention: a shorter, non-normalized bar leaves space at the top, not the
   bottom) with a small gap between bars. Population stacking order is
   **global** (the same population always occupies the same relative band in
   every bar); double-clicking a column re-ranks all rows by that column's
   values, putting its largest population at the bottom everywhere.
8. **`composition_col_weights`** — carries `DatasetCollection.calc_population`'s
   true per-group `n_cells` onto the widget, so the `PROP`/`COUNTS` toggle can
   show *real* group-size differences (taller bar for a bigger dataset) even
   though the displayed matrix itself holds proportions (which always sum to 1
   per group, discarding absolute size on their own).
9. **Row labels** replace the old click-to-sort legend — shown next to the
   leftmost bar, colored to match their segment, double-click to reorder (same
   mechanism as #7). Labels are hidden when their segment is too short to fit
   one line of text, and reveal themselves as you zoom in on rows (rather than
   always showing, however small, which cut off badly for small populations
   or in `COUNTS` mode).
10. **Row dendrogram** — dynamically positioned from the rightmost bar's actual
    (non-uniform) segment geometry, so clusters of co-regulated populations
    across datasets are visible, the same way a regular `Clustergram` row
    dendrogram would show them. Recomputes on reorder/normalize/weight changes.
    Column dendrogram trapezoids account for the gap between bars. Both axes'
    trapezoids animate their shape on resize.
11. **Cross-bar hover highlight** — hovering a segment, a row label, or a
    column label dims every other population's/group's segments across all
    bars (after a short dwell delay), so comparing shares across groups is
    easy.
12. **Dendrogram hover/click highlight** — hovering or clicking a dendrogram
    trapezoid dims every row/column *not* covered by it — in both
    `Clustergram` and `Composition`.
13. **Vertical-only zoom** — columns/datasets always stay fully visible; only
    row (population) detail zooms.
14. **Tooltip** — surfaces the group's total cell count alongside the hovered
    cell's value.

**Control panel (both widgets)**
15. Every button (row/col reorder `ini`/`sum`/`var`/`clust`, new toggles)
    restyled: capitalized text, no border/background — state shown purely by
    text color (blue = active, gray = inactive). Non-clickable axis-name
    labels get a colon (`Cell:`, `Gene:`, `DSET:`, `POP:`) and stay black,
    fixed-width so both rows align, with a small margin from the panel border.

**Interaction robustness** (found and fixed while building the above)
16. A hover-highlight (composition bars, dendrogram, or categorical attribute
    tiles) could get stuck showing its last state after the mouse left the
    widget, if a pending delayed-highlight timer fired after the fact — fixed
    with a failsafe clear on pointer-leave plus proactive cross-clearing
    between adjacent hover regions.
17. The widget container's CSS width didn't account for the deck.gl canvas's
    own rendering buffer (`width + 100`px, mirrored for height but not width),
    so content at the far right edge — the row dendrogram — could fall
    outside the box the browser tracked mouse events against, causing
    mouse-out to misbehave specifically there.
18. Categorical attribute tile hover now has the same short dwell delay as
    every other hover-highlight, for consistency.

## API

```python
import celldega as dega

# 1. Dot-plot Clustergram: mean expression (color) + fraction expressing (size)
setc = dega.SetCollection(adata, set_col="leiden")
setc.calc_signature(adata, modality_name="expression", aggregate="mean", normalization=None)
setc.calc_signature(adata, modality_name="fraction_expressing", aggregate="fraction")

mat = dega.clust.Matrix(collection=setc, color_by="expression", size_by="fraction_expressing")
mat.clust()

cg = dega.viz.Clustergram(matrix=mat, width=550, height=550)  # viz_mode defaults to "dotplot"
cg.dot_size_encoded = False   # (dotplot only) force full tile, ignore the fraction channel

# 2. Composition: cell-type proportions across datasets, with true per-dataset counts
dset = dega.DatasetCollection(adata, dataset_col="sample_id", obs_columns=["condition"])
dset.calc_population(adata, category="cell_type")

comp = dega.viz.Composition(
    dset, category="cell_type", group_attrs=["condition"], width=650, height=420,
)  # colors come from dset alone -- calc_population already carried adata's palette onto it
comp                                  # stacked bars; PROP/COUNTS + reorder in the control panel
comp.composition_normalized = False   # animate to raw-count heights (uses dset.obs["n_cells"])
```

## Details

- **Body encoding pipeline** (`js/matrix/mat_data.js`, `js/deck-gl/matrix/mat_layer.js`,
  `mat_shaders.js`): a per-instance `instanceRadius` shader attribute (default
  `1.0` = full tile, so classic heatmap is unchanged) scales each square around its
  center. `encode_point()` maps `viz_mode` to a `(color/alpha, size_scale)` pair per
  cell — `heatmap`: alpha ∝ value, full size; `dotplot`: alpha from the main
  matrix, size from the secondary (or forced to `1` when `dot_size_encoded=False`).
- **Composition layout** (`js/matrix/composition_data.js`, `composition_layer.js`):
  segment height = share of column × bar height; bar height = full when normalized,
  else scaled by `composition_col_weights` (falling back to the column's own sum —
  algebraically identical to the old un-weighted formula when no weights are
  given, so a raw-count matrix behaves the same either way). Global row order comes
  from the shared `orders.row` state; double-click reorder (`custom_label_reorder`,
  shared with regular `Clustergram`) sets a `'custom'` order ranked so the highest
  value gets the largest rank, which this module maps to the bottom.
- **Zoom-axis lock** (`js/deck-gl/matrix/views.js`, `on_view_state_change.js`): a
  permanent `viz_mode === 'composition'` branch alongside the existing shape-driven
  major/minor-axis system (which is a *temporary* aspect-ratio delay that always
  unlocks to both axes — reused as-is it would let X zoom eventually anyway).
- **Dendrogram** (`js/matrix/dendro.js`, `js/deck-gl/matrix/dendro_layers.js`):
  in composition mode, row-axis leaf spans come from that row's actual segment in
  the rightmost bar (`get_composition_layout`) rather than a uniform per-row slot;
  column-axis leaf spans account for the same small gap `build_composition_layout`
  renders between bars. Hover/click highlight is a single `dendro_highlight_alpha_factor(row, col)`
  resolver shared by `mat_layer.js`'s and `composition_layer.js`'s `getFillColor`
  accessors. Trapezoid shape transitions (`transitions.getPolygon`) are safe now
  that the decorative stroke sublayer is gone (`stroked: false`) — a fixed
  3-vertex triangle's fill has unambiguous per-vertex interpolation, whereas the
  stroke's own path-based transition was the source of an earlier animation glitch.
- **Row label fit** (`js/matrix/composition_data.js`): a row label is hidden
  whenever the leftmost bar's segment for that row is shorter than one line of
  text at the current zoom; the fit ratio is zoom-dependent (row labels use a
  fixed screen-pixel size, not scaled by zoom, unlike everything else) so
  zooming in on rows reveals labels that didn't fit before.
- **Hover-highlight robustness** (`composition_layer.js`, `dendro_layers.js`,
  `cat_layers.js`, `matrix_viz.js`): every delayed hover-highlight cancels its
  own pending timer on any subsequent hover event; a `pointerleave` listener on
  the widget root force-clears all three (composition, dendrogram, categorical)
  as a failsafe; hovering the matrix body or a row/col label proactively clears
  the dendrogram highlight, since they can't legitimately be active at once.
- **`composition_col_weights`** (`widget.py`, `composition_data.js`): read from
  `calc_population`'s modality `obs["n_cells"]` (always present, independent of
  `output` mode) or the collection's own `obs`; keyed by group name, looked up per
  column via `col_nodes[c].name`.
- **Control-panel buttons** (`js/ui/text_buttons.js`, `ui_containers.js`): a shared
  `apply_state_button_style`/`make_text_toggle_group` primitive replaces four
  previously independent hand-rolled pill-button implementations for the buttons
  touched here; axis labels use a fixed-width flex box (not text-width-dependent)
  so reorder buttons in both rows start at the same x position regardless of
  entity-name length.

## Scope

**Not included** (flagged during review, kept out of scope): a per-population
"survey" bar chart for the row axis (mirroring the column-attribute bars), and
a height-vs-opacity segment-encoding toggle for `Composition` (considered, then
removed — that's what regular `Clustergram` heatmap mode already covers).

## Files

**Python**
- `src/celldega/viz/widget.py` (+333) — `viz_mode` (+ validator restricting
  `"composition"` to `Composition` and rejecting `"size"`), `dot_size_encoded`,
  `composition_normalized`/`composition_colors`/`composition_col_weights` traits;
  `Composition` class; `_composition_matrix_inputs`.
- `src/celldega/clust/matrix.py` (+133) — `Matrix.set_dot_matrix` (+ `AnnData`
  transpose fix), `dot_mat` parquet export, `Matrix(collection=..., color_by=...,
  size_by=...)`.
- `src/celldega/set/collection.py` (+56) — `calc_signature` required
  `modality_name`, `calc_population` color-palette carry-through.
- `src/celldega/viz/__init__.py` (+3), `src/celldega/__init__.py` (+3) — export
  `Composition`.

**JS**
- `js/matrix/composition_data.js` (+289), `js/deck-gl/matrix/composition_layer.js`
  (+272), `composition_shaders.js` (+55) — composition layout engine + deck.gl layer.
- `js/deck-gl/matrix/dendro_layers.js` (+222), `js/matrix/dendro.js` (+63) —
  composition-aware trapezoid positioning, hover/click highlight, resize animation.
- `js/viz/matrix_viz.js` (+205) — mode switching, trait listeners, zoom-lock/hover wiring.
- `js/ui/ui_containers.js` (+141), `js/ui/text_buttons.js` (+149) — shared button
  primitives, control-panel layout/toggles/spacing.
- `js/deck-gl/matrix/label_layers.js` (+154), `js/deck-gl/matrix/cat_layers.js`
  (+130) — row-label position/visibility/hover, categorical hover delay.
- `js/matrix/mat_data.js` (+83), `js/matrix/set_constants.js` (+49),
  `js/deck-gl/matrix/mat_layer.js` (+84), `mat_shaders.js` (+6) — dotplot
  encoding pipeline, container width fix.
- `js/deck-gl/matrix/on_view_state_change.js` (+33), `views.js` (+10) —
  vertical-only zoom lock, zoom-dependent row-label fit.
- `js/deck-gl/matrix/matrix_tooltip.js` (+29), `matrix_layers.js` (+30),
  `js/matrix/attr_state.js` (+13), `js/ui/attribute_labels.js` (+21) — tooltip,
  layer-filter fix, shared update-trigger key.
- `js/read_parquet/network_from_dega_files.js` (+12), `network_from_parquet.js`
  (+9), `js/celldega.js` (+3/-2) — `dot_mat` plumbing (comm and bucket-file paths).

**Tests**
- `tests/unit/test_viz/test_widget.py` (+93) — `viz_mode` validation
  (`"composition"`/`"size"` rejection), `dot_size_encoded` default,
  `Composition` subclass/defaults, `composition_col_weights`.
- `tests/unit/test_clust/test_matrix.py` (+82) — `set_dot_matrix` `AnnData`
  alignment, `Matrix(collection=..., color_by=..., size_by=...)`.
- `tests/unit/test_schema/test_set_collection.py` (+10) — `modality_name`
  required in `calc_signature`.
- Full suite: **260 passed**.

**Notebooks**
- `notebooks/Clustergram_Visual_Encodings.ipynb` (+484) — heatmap/dotplot walkthrough.
- `notebooks/Composition_Population_Proportions.ipynb` (+567) — `Composition`
  walkthrough, including the `n_cells` → `composition_col_weights` COUNTS-mode demo.

## Tests

Covers: `viz_mode` validation (rejects `"composition"` off `Composition`, rejects
`"size"` entirely) and `dot_size_encoded` toggling on `Clustergram`; `Composition`
as a real `Clustergram` subclass with correct `viz_mode`/`composition_normalized`
defaults; `composition_col_weights` populated from `DatasetCollection.calc_population`'s
`n_cells` (and empty for a bare `DataFrame` with no such metadata); `set_dot_matrix`
correctly aligns `AnnData` input by name; `Matrix(collection=..., color_by=...,
size_by=...)` builds and wires up correctly, including the `dot_plot=` alias and
error cases; `calc_signature` requires `modality_name`. Full suite: 260 passed.

cornhundred (commented):
Will merge in changes and publish v0.20.0

PR #327, opened 2026-08-04
Manual landmark weight

cornhundred:


cornhundred (commented):
Merging into main v0.22.0

PR #329, opened 2026-08-05
fixing yearbook object array bug

cornhundred:


cornhundred:
## DEGA-593: Fix Yearbook selection, coloring, and layout bugs

Brings the `Yearbook` widget's gene/cluster interactions in line with
`Landscape`, and removes dead space below the portrait grid. All functional
fixes are front-end (JS); no Python API changes. Version bump 0.23.0 → 0.23.1.

### What was broken & why

| Symptom | Root cause | Fix |
|---|---|---|
| Every portrait showed the same cell | Centering read `cell_scatter_data_objects`, a buffer no longer populated after centroids moved to flat typed arrays → all lookups missed → fell back to landscape center | Read the flat `cell_scatter_data` buffer via `cell_name_to_index_map` |
| Cell polygons vanished on gene select | `get_path_color` filtered polygons by `selected_cats`, which holds the *gene name* in gene mode → no cluster matched → alpha 0 for all | Apply the cluster filter only to real cluster categories |
| Centroids didn't turn red on gene select | 2D scatterplot path cloned the layer id without rebuilding the color buffer | Rebuild scatterplot data via `refresh_cell_layer_data` |
| Querying a gene didn't focus its transcripts | Query assigned `selected_genes` directly, leaving `selected_gene_ids` (read by the trx layer) stale | Route through new `force_set_selected_genes`, which syncs the Set |
| Clicking a bar didn't update the highlight | Rect-fill alpha (re-render) and group opacity (click) fought | Unify both paths on group opacity via `get_bar_highlight_opacity` |
| Dead space under the portrait grid | Canvas/root used a fixed height; width-limited grids are shorter | Size to actual grid height via `get_grid_pixel_height` |

### How to test
1. Open a `Yearbook` with a multi-cell selection → portraits show **distinct** cells.
2. Query a gene → centroids turn **red** (opacity ∝ expression), polygons stay cluster-colored, the queried gene's **transcripts are emphasized**, and the gene bar graph scrolls to + highlights that gene.
3. Click gene/cluster bars → clicked bar full opacity + bold, others dimmed.
4. Use a wide grid (e.g. 6×3) → no gap below the last row.

### Tests & CI
- New JS unit tests (114 total, all green): `yearbook_portrait_centers`,
  `path_layer_gene_mode`, `bar_highlight_opacity`, `force_set_selected_genes`,
  `yearbook_grid_height`.
- Quality gate green locally: `eslint` (0 errors), `ruff check src/`,
  `ruff format --check src/`, `prettier --check`. Includes a small
  `import/order` fix (blank line between import groups) so `npm run lint:js`
  passes.

### ⚠️ Release note
The widget loads `celldega.js` from jsdelivr pinned to the package version, so
these fixes only reach users once the built bundle is published to npm for
`0.23.1`.

